PR #28, opened 2023-04-21
feat: acl page improvements

shishovKA:
- turn on subjectName filter (for tables: Responsibles & Object Permissions)
- change inherited icon + add popover tip


packages/ui/package-lock.json
```diff
@@ -6676,19 +6676,6 @@
...
-      "integrity": "sha512-ihpxHaFc2MIGsXEnjdKy21jQrs/sV/QOCSdTmAwIjKXZ+9qKeYAbpou9HPzGwNJSu7bJMGnrLNxh6jodaYKUww=="
-    },
-    "@ytsaurus/javascript-wrapper": {
-      "version": "0.2.1",
-      "resolved": "https://registry.npmjs.org/@ytsaurus/javascript-wrapper/-/javascript-wrapper-0.2.1.tgz",
-      "integrity": "sha512-h884FY9pSb/Y7BIM/BV0gXlKazqTOOkihHuLIS+eijWyYP1Sebx+/neXHy+ib+CMepeeYoG3AOT9bAewL4n7Wg==",
-      "requires": {
-        "axios": "^0.21.4"
-      }
-    },
```
ma-efremoff: I am sure we don't need this changes and we don't need last commit which revert it.

packages/ui/src/ui/components/ACL/ACL-connect-helpers.js
```diff
@@ -67,6 +68,7 @@ const makeAclMapStateToProps = (inputIdmKind) => {
             disableInheritanceResponsible,
         } = state.acl[idmKind];
 
+        const hasApprovers = getHasApprovers(state, idmKind);
         const approvers = getApproversOrderedByInheritanceAndSubject(state, idmKind);
```
ma-efremoff, line 71: `approvers` contains result of filtering, it is better to reflect this in the variable name. `getApproversFiltered`/`getApproversFilteredAndOrdered` would be more descriptive, without this, it is not clear why we need to call `getHasApprovers` here.

ma-efremoff (approved):

PR #36, opened 2023-04-27
fix(TabletCellBundle): better layout for meta-table [YTFRONT-3716]

ma-efremoff:


packages/ui/src/ui/pages/tablet_cell_bundles/bundle/BundleGeneralMeta.scss
```diff
@@ -3,8 +3,10 @@
         grid-gap: 6px 27px;
     }
 
+    $progressWith: 200px;
```
shishovKA, line 6: looks like a typo in word: $progressWi**d**th

shishovKA (approved):

PR #39, opened 2023-05-02
feat: ACL add object permissions own filters [YTFRONT-3720]

shishovKA:
On ACL page subject filter was splited on two separate filters: for _Responsibles_ table and for _Object permissions_ table.
Also, permissions filter was added for _Object permissions_ table.


packages/ui/src/ui/store/selectors/acl.js
```diff
@@ -60,9 +64,23 @@ function FilterBySubject(items, subjectFilter) {
...
-    [getAllObjectPermissionsWithSplittedSubjects, getObjectSubjects],
-    FilterBySubject,
+    [
+        getAllObjectPermissionsWithSplittedSubjects,
+        getObjectSubjectFilter,
+        getObjectPermissionsFilter,
+    ],
+    (items, subjectFilter, permissionsFilter) => {
+        const buf = FilterBySubject(items, subjectFilter);
+        return FilterByPermissions(buf, permissionsFilter);
```
ma-efremoff, line 82: It is possible to iterate over all items only once but you need to prepare a predicate function, we have some helpers in predicates.ts, please look up at `concatByAnd`. Also there is no need to filter items if filters are empty.

shishovKA: fixed 👍🏻

packages/ui/src/ui/constants/acl.ts
```diff
@@ -7,9 +7,7 @@ export const LOAD_DATA = createActionTypes(PREFIX + 'LOAD_DATA');
 export const DELETE_PERMISSION = createActionTypes(PREFIX + 'DELETE_PERMISSION');
 export const REQUEST_PERMISSION = createActionTypes(PREFIX + 'REQUEST_PERMISSION');
 export const UPDATE_ACL = createActionTypes(PREFIX + 'UPDATE_ACL');
-export const CHANGE_OBJECT_SUBJECT = PREFIX + 'CHANGE_OBJECT_SUBJECT';
-export const CHANGE_COLUMNS_SUBJECT = PREFIX + 'CHANGE_COLUMNS_SUBJECT';
-export const CHANGE_COLUMNS_COLUMNS = PREFIX + 'CHANGE_COLUMNS_COLUMNS';
+export const CHANGE_FILTERS = `${PREFIX}CHANGE_FILTERS` as const;
```
ma-efremoff, line 10: I think it better to use a name linked with ALC, like `ACL_CHANGE_FILTERS`.

ma-efremoff (approved):

PR #48, opened 2023-05-16
feat(Navigation/Consumer): Model is changed to 'many-to-many' [YTFRONT-3327] 

ma-efremoff:


packages/ui/src/ui/store/actions/navigation/content/table/table.js
```diff
@@ -367,6 +367,7 @@ function restoreColumns(state) {
 
 // TODO: consider switching back for <link rel="preload"> once it's supported in all major browsers
 function preloadTableFont(fontFamily) {
+    console.log({fontFamily});
```
shishovKA, line 370: Could be removed (if was used for debug)

shishovKA (approved):

PR #51, opened 2023-05-19
feat(System/Nodes): add 'Node type' filter [YTFRONT-3163]

ma-efremoff:


shishovKA (approved):

packages/ui/src/ui/pages/system/Nodes/Nodes.js
```diff
@@ -39,17 +50,26 @@ class Nodes extends Component {
...
+        const {racks, rackGroups, nodeType, containerWidth} = this.props;
+        const headingCN = cn('elements-heading')({
             size: 's',
             overview: 'yes',
         });
 
+        if (_.isEmpty(racks) && _.isEmpty(rackGroups)) {
+            return (
+                <NoContent
+                    warning={`There are no nodes for ${format.ReadableField(nodeType?.join(','))}`}
```
shishovKA, line 62: Maybe change warning text to:
`There are no {nodeType}` because word 'nodes' duplicates

image

also it is better to add space after coma to `nodeType?.join(', ')`, because when 2 or more node types selected - warning text will be stuck together.
Now warning text might look like:
`There are no nodes for Chaos node,tablet nodes`

shishovKA (approved):

PR #54, opened 2023-05-25
feat(AttributesEditor): edit expiration_time and expiration_timout [YTFRONT-3665]

ma-efremoff:


packages/ui/src/ui/pages/navigation/modals/AttributesEditor.tsx
```diff
@@ -44,11 +45,37 @@ import {docsUrl} from '../../../config';
...
+        in_memory_mode: InMemoryModeType | '';
+    };
+    storage: {
+        path: Array<{title: string}>;
+        optimize_for: 'lookup';
+        compression_codec: Array<string>;
+        erasure_codec: string;
+        replication_factor: number;
+
+        expiration_time: {from?: string | number};
```
shishovKA, line 66: I think it is better to use `{to: string | number}` DatePicker prop for `expiration_time` value.

> In meaning: date when smth will be deleted

shishovKA (changes requested):

packages/ui/src/ui/pages/navigation/modals/AttributesEditor.tsx
shishovKA: It is ok to use `from` prop here, but Datepicker also have `to` prop which works same - just looks more suitable to describe `expiration_time`

For me it is not an issue, just my improvement thoughts 👌🏻

ma-efremoff: `to` is defined only for ranges:
image

packages/ui/src/ui/pages/navigation/modals/AttributesEditor.tsx
```diff
@@ -44,11 +45,37 @@ import {docsUrl} from '../../../config';
...
+        primary_medium: string;
+        tablet_cell_bundle: string;
+        in_memory_mode: InMemoryModeType | '';
+
+        expiration_time: {from?: string | number};
+        expiration_timeout: {value: number};
+    };
+    storage: {
+        path: Array<{title: string}>;
+        optimize_for: 'lookup';
```
shishovKA, line 64: 'lookup' | 'scan' ?

shishovKA (approved):

PR #65, opened 2023-06-09
fix: ACL page tables styles [YTFRONT-3758]

shishovKA:
- reworked all ACL tables with DataTableYT component
- adjusted column width settings (tried to avoid overlapping and get almost same width for similar columns on this page)

### DIFF
--- a/packages/ui/src/ui/components/ACL/ACL.scss
+++ b/packages/ui/src/ui/components/ACL/ACL.scss
@@ -4,7 +4,7 @@
     position: relative;
 
     display: grid;
-    grid-template-columns: minmax(0, 1fr) 300px;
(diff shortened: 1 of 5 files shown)


packages/ui/src/ui/components/ACL/ACL.scss
```diff
@@ -77,16 +82,16 @@
     }
 
     .elements-table {
```
ma-efremoff, line 84: It looks like ACL.js don't use ElementsTable anymore. Is it really need here?

shishovKA: Yes, we don't need `.elements-table` style rules here anymore, I am going to remove them from here and update farm stands 👍🏻

ma-efremoff (approved):

PR #66, opened 2023-06-14
feat: remote copy modal -> suggest transfer_* pool if exists [YTFRONT-3511]

shishovKA:


packages/ui/src/ui/components/Dialog/controls/PoolSuggestControl/PoolSuggestControl.tsx
```diff
@@ -20,6 +20,7 @@ type Props = DialogControlProps<string> & {
     cluster?: string;
     // If not defined then default pool tree should be used
     poolTree?: string;
+    setOnLoadPoolsValue?: (poolNames: string[]) => string;
```
ma-efremoff, line 23: I am not sure, but may be it is possible to make up a better name:
`calculateValueFromLoaded`/`calculateValueFromLoadedItems`/`recalculateValueOnLoad`/...
```ts
calculateValueFromLoaded({loadedPoolNames}: {loadedPoolNames: Array<string>})
```
What do you think?
And may be it is better to pass `cluster` as a parameter to this callback instead of using closures.

packages/ui/src/ui/pages/navigation/modals/RemoteCopyModal.tsx
```diff
@@ -186,6 +185,23 @@ function RemoteCopyModal() {
...
+                                    true,
+                                ),
+                            )}
+                        </>
+                    ),
+                    visibilityCondition: {
+                        when: 'dstCluster',
+                        isActive: (value) => Boolean(value),
+                    },
+                    required: true,
```
ma-efremoff, line 204: According to the documentation
https://ytsaurus.tech/docs/en/user-guide/data-processing/operations/operations-options#common_options
the parameter might be omitted.

packages/ui/src/ui/pages/navigation/modals/RemoteCopyModal.tsx
```diff
@@ -186,6 +185,23 @@ function RemoteCopyModal() {
...
+                            {docsUrl(
+                                makeLink(
+                                    UIFactory.docsUrls['operations:remote_copy'],
+                                    'limits',
+                                    true,
+                                ),
+                            )}
+                        </>
+                    ),
+                    visibilityCondition: {
```
ma-efremoff, line 200: I think If the field is not marked as required we don't need this condition, do we?

packages/ui/src/ui/components/Dialog/controls/PoolSuggestControl/PoolSuggestControl.tsx
shishovKA: > I am not sure, but may be it is possible to make up a better name: `calculateValueFromLoaded`/`calculateValueFromLoadedItems`/`recalculateValueOnLoad`/...
> 
> ```ts
> calculateValueFromLoaded({loadedPoolNames}: {loadedPoolNames: Array<string>})
> ```
> 
> What do you think? And may be it is better to pass `cluster` as a parameter to this callback instead of using closures.

I'll try to find better name 👍🏻

We pass _destination_ cluster name to `PoolSuggestControl` prop, and if we want to return _current_ cluster name as callback parameter we could:
 - connect `PoolSuggestControl` to the store, to get current cluster name inside control component
 - or add one more prop `currentCluster` to return it back with callback
I think these two options are more complicated and hide part of the logic inside control 🤔

ma-efremoff (approved):

PR #85, opened 2023-07-13
fix: PoolEditorDialog weight editing [YTFRONT-3748]

shishovKA:


packages/ui/src/ui/utils/scheduling/pool-child.js
```diff
@@ -44,124 +44,156 @@ function preparePoolChildResource(data, type, treeResources, resource) {
 }
 
 export function updatePoolChild(data, cypressData, type, treeResources, opts) {
-    const attributes = data.attributes;
-    const cypressAttributes = ypath.getAttributes(cypressData);
+    try {
```
shishovKA, line 47: just wrapped with try{} catch{}
other diffs are -> indentation

packages/ui/src/ui/common/thor/ypath.js
```diff
@@ -79,9 +79,16 @@ thorYPath.getBoolean = function (node, path) {
...
 thorYPath.getNumber = function (node, path, defaultValue) {
-    const value = thorYPath.get(node, path);
-
-    return convertToNumber(value, defaultValue);
+    try {
+        const value = thorYPath.get(node, path);
+        return convertToNumber(value, defaultValue);
+    } catch (e) {
+        const error = {
+            message: `thorYPath: failed to convert field with path: "${path}".`,
```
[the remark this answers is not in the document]
ma-efremoff: Will be the error displayed correctly if we add inner_errors to the caught error without wrapping to a new object?

packages/ui/src/ui/utils/scheduling/pool-child.js
```diff
@@ -44,124 +44,156 @@ function preparePoolChildResource(data, type, treeResources, resource) {
...
+        preparePoolChildResource(data, type, treeResources, 'cpu');
+        preparePoolChildResource(data, type, treeResources, 'user_memory');
+        preparePoolChildResource(data, type, treeResources, 'gpu');
+        preparePoolChildResource(data, type, treeResources, 'user_slots');
+
+        return data;
+    } catch (e) {
+        const error = {
+            message: `An error occured while parsing pool "${data.name}" data:`,
+            inner_errors: [e],
```
ma-efremoff, line 195: The same as above. What about adding `inner_errors` instead of wrapping by an object?

ma-efremoff, line 195: The same as above. What about adding `inner_errors` instead of wrapping by an object?

packages/ui/src/ui/utils/scheduling/scheduling.ts
```diff
@@ -216,7 +216,10 @@ export function getInitialValues(editItem: PoolInfo | undefined, allowedSources:
                 : undefined,
             parent: editItem?.parent,
             mode: editItem?.mode,
-            weight: ypath.getValue(editItem, '/cypressAttributes/weight') || '',
+            weight: {
+                value: ypath.getValue(editItem, '/cypressAttributes/weight') || undefined,
+                error: undefined,
```
ma-efremoff, line 221: As I remember, the error field is optional. Might it be omitted here?

shishovKA:
Scheduler error now catches and looks:

Screenshot 2023-07-12 at 13 40 33

packages/ui/src/ui/common/thor/ypath.js
```diff
@@ -79,9 +79,17 @@ thorYPath.getBoolean = function (node, path) {
...
-
-    return convertToNumber(value, defaultValue);
+    try {
+        const value = thorYPath.get(node, path);
+        return convertToNumber(value, defaultValue);
+    } catch (e) {
+        const curError = {
+            message: `thorYPath.getNumber: failed to convert field with path: "${path}".`,
+        };
+        if (e?.inner_errors) e.inner_errors.push(curError);
```
ma-efremoff, line 89: The action of adding an item to `inner_errors` should be exception safe. I think it is a bad idea to rely on that type of inner_errors might be defined only as `Array` or `undefined`. It is better to check it and handle it properly, probably, we need helper function.

shishovKA: i have added a helper function `appendInnerErrors`:

you could use this compare link to watch recent chages: https://github.com/ytsaurus/ytsaurus-ui/compare/4e9c258381505e3e485be3ab3df56c35565dc8fb..7852283a95e27c204df8bde4358e56f8a80ed474

packages/ui/src/ui/utils/errors/index.ts
```diff
@@ -20,3 +20,12 @@ export function getErrorWithCode(errors: YTError[], code: number): YTError | und
 export function getPermissionDeniedError(error: YTError): YTError | undefined {
     return getErrorWithCode([error], yt.codes.PERMISSION_DENIED);
 }
+
+export function appendInnerErrors(targetErr: any, innerErr: any) {
+    if (targetErr.inner_errors && Array.isArray(targetErr.inner_errors)) {
```
ma-efremoff, line 25:
```ts
if (!targetErr) {
  targetErr = new Error('Unexpected behavior: targetErr is undefined');
}

if (!targetErr.inner_errors) {
  targetErr.inner_errors = [innerErr];
  return targetErr;
}

if (Array.isArray(targetErr.inner_errors)) {
  targetErr.inner_errors.push(innerErr);
} else {
  targetErr.inner_errors = [targetErr.inner_errors, innerErr];
}
return targetErr;
```

packages/ui/src/ui/utils/errors/index.ts
```diff
@@ -20,3 +20,12 @@ export function getErrorWithCode(errors: YTError[], code: number): YTError | und
 export function getPermissionDeniedError(error: YTError): YTError | undefined {
     return getErrorWithCode([error], yt.codes.PERMISSION_DENIED);
 }
+
+export function appendInnerErrors(targetErr: any, innerErr: any) {
```
ma-efremoff, line 24: `any` is OK for `targetErr`, but I think `YTError` fits better for `innerErr`.

ma-efremoff (approved):

PR #95, opened 2023-07-18
fix: QT format decimal number results [YTFRONT-3782]

shishovKA:
Arcanum link: https://nda.ya.ru/t/x-CECAO46VmeKP
Query to test: https://nda.ya.ru/t/dU3bnvVn6VsBSY


packages/ui/src/ui/pages/query-tracker/module/query_result/utils/format.ts
```diff
@@ -14,7 +13,6 @@ const defaultSettings = {
     maxListSize: 50,
     maxStringSize: undefined as number | undefined,
     showYsonAs: 'yson',
-    customNumberFormatter: hammer.format.Number, //getNumberFormatter(),
```
shishovKA, line 17: removing this field -> we let unipika format values itself

ma-efremoff (approved):

PR #96, opened 2023-07-20
fix: PoolEditorDialog weight field remove request when value was not changed [YTFRONT-3748]

shishovKA:
https://nda.ya.ru/t/n-Iq4RjF6W2tVW

### DIFF
--- a/packages/ui/src/ui/common/thor/utils.ts
+++ b/packages/ui/src/ui/common/thor/utils.ts
@@ -1,0 +1,36 @@
+//@ts-ignore
+import ypath from '@ytsaurus/interface-helpers/lib/ypath';
+import {appendInnerErrors} from '../../utils/errors';
+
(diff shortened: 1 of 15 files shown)


packages/ui/src/ui/common/thor/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+export function convertToNumber<T extends number | undefined>(
+    value: any,
+    defaultValue?: T,
+): number | T {
+    if (value === null) return defaultValue as T;
+    const res = Number(value);
+    return isNaN(res) ? (defaultValue as T) : res;
+}
+
+export const getNumber = function <T extends number | undefined>(
```
ma-efremoff, line 14: It is better to have some tests for `getNumber` too. It looks like an exception will be never thrown from the function.

PR #98, opened 2023-07-25
feat(QT): add QT settings override with UI [YTFRONT-3790]

shishovKA:
https://nda.ya.ru/t/ERLAgHms6Wifso

### DIFF
--- a/packages/ui/src/ui/containers/SettingsPanel/settings-description.tsx
+++ b/packages/ui/src/ui/containers/SettingsPanel/settings-description.tsx
@@ -157,6 +157,25 @@ function useSettings(cluster: string, isAdmin: boolean): Array<SettingsPage> {
                             }
                         />,
                     ),
+                makeItem(
+                    'Query Tracker Stage',
+                    'top',
+                    <SettingsMenuInput
+                        placeholder="Enter stage..."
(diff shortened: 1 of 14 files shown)


packages/ui/src/ui/store/index.ts
```diff
@@ -32,3 +32,4 @@ export function createAppStore() {
 
 export type AppBrowserHistory = ReturnType<typeof createAppStore>['history'];
 export type StoreType = ReturnType<typeof createAppStore>['store'];
+export type AppThunkDispatch = ThunkDispatch<RootState, any, AnyAction>;
```
ma-efremoff: Probably it is better to write separate function like `useAppDispatch`/`useThunkDispatch` and use it when we need to know type of returned value from an async action.

packages/ui/src/ui/containers/SettingsPanel/settings-description.tsx
```diff
@@ -457,6 +458,32 @@ function useSettings(cluster: string, isAdmin: boolean): Array<SettingsPage> {
                 ),
             ]),
         ),
+        isAdmin &&
+            makePage(
```
ma-efremoff: If we don't want to allow edit the settings for regular users it is better to put them under `Development` section of settings.

ma-efremoff (approved):

PR #112, opened 2023-07-28
some small tasks

shishovKA:


packages/ui/src/ui/utils/index.ts
```diff
@@ -279,6 +279,12 @@ export function parseSortState(stringSortState: string) {
     return parsed;
 }
 
+export function parseSortStateNoFlags(stringSortState: string) {
+    const values = _.map(stringSortState.split(','), (value) => value.split(/-(.*)/s));
```
ma-efremoff, line 283: Will it break anything if we use the regular expression from `parseSortState`?
It is better to leave only one function if it is possible.

shishovKA: This method:
https://github.com/ytsaurus/ytsaurus-ui/blob/0a4ba7c15730854c9b2d7224cf7bc87fefb75a90/packages/ui/src/ui/utils/index.ts#L269

works bad with query values like: `order-undefined-asc`, `order-asc-undefined`
it splits them as: `[order, undefined, asc]`, `[order, asc, undefined]` and causes wrong url-mapping which brokes our sort state and icons

Also it has additional logic with flags for old-style sort states.
We need these values be splitted  like: `[order, undefined-asc]` and `[order, asc-undefined]`

ma-efremoff (approved):

PR #113, opened 2023-07-31
fix: replace //sys/proxies with //sys/http_proxies [YTFRONT-3799]

ma-efremoff:


packages/ui/ytsaurus-ui-empty.tgz
shishovKA, line 1: Was this file deleted by mistake? or it's OK?

shishovKA (approved):

PR #115, opened 2023-08-03
fixes for BundleEditor

ma-efremoff:
https://nda.ya.ru/t/OqD3q0Qk6ZBQGB

### DIFF
--- a/packages/ui/src/ui/pages/tablet_cell_bundles/bundles/BundleEditorDialog/BundleEditorDialog.tsx
+++ b/packages/ui/src/ui/pages/tablet_cell_bundles/bundles/BundleEditorDialog/BundleEditorDialog.tsx
@@ -19,7 +19,10 @@ import {
     hideTabletCellBundleEditor,
     setBundleEditorController,
 } from '../../../../store/actions/tablet_cell_bundles/tablet-cell-bundle-editor';
-import {getBundleEditorData} from '../../../../store/selectors/tablet_cell_bundles';
+import {
+    getBundleEditorData,
+    getTabletBundlesWriteableByName,
+} from '../../../../store/selectors/tablet_cell_bundles';
 import {BundleResourceGuarantee} from '../../../../store/reducers/tablet_cell_bundles';
 import {getTabletCellBundleEditorState} from '../../../../store/selectors/tablet_cell_bundles/tablet-cell-bundle-editor';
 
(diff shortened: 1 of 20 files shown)


packages/ui/src/ui/pages/tablet_cell_bundles/bundles/BundleEditorDialog/BundleEditorDialog.tsx
```diff
@@ -318,6 +333,15 @@ export function BundleEditorDialog() {
                     },
                 },
                 validator: simpleBundleValidate,
+                tooltip: (
```
shishovKA, line 336: If I'm not mistaken, now we will never see these fields when BC is off?
Because editing is not permitted for BC=off ((

ma-efremoff: `renderSimpleResourceTab` is used only when BC == 'off' and current user is allowed to write //sys/tablet_cell_bundles/{bundleName} node. So to see the message a user should have write access to a bundle.

packages/ui/src/ui/pages/tablet_cell_bundles/bundles/BundlesTable.tsx
```diff
@@ -514,6 +517,7 @@ export type ReduxProps = {
     columns: Array<keyof typeof Columns>;
     activeBundleLink(cluster: string, bundle: string, enable_bundle_controller?: boolean): string;
     bundleDashboardUrl?(cluster: string, bundle: string): string | undefined;
+    writeableByName?: {get: (bundleName: string) => boolean | undefined};
```
shishovKA, line 520: Is it possible to define field type as `Map<string, boolean>`?

shishovKA (approved):

packages/ui/src/ui/pages/tablet_cell_bundles/bundles/BundlesTable.tsx
ma-efremoff: Yes, it is allowed to pass instances of Map<string, boolean> as a value for the property. It will be the same if we rewrite the type as 
```ts
writeableByName?: Pick<Map<string, boolean>, 'get'>;
```
If you think it is more clear I will replace the declaration.

shishovKA: both variants are ok for me, I just was worried about picking only one method instead of full Map
but then i`ve understood the reason 👍🏻

PR #129, opened 2023-08-23
feat: QT add errors to vs-code editor

shishovKA:
Arcadia PR: https://nda.ya.ru/t/e_iXTU436cWYnK

Features:

- parse query inner_errors to get any vs-code coordinates and decorate them with red curly-underlines and line background [example-1](https://nda.ya.ru/t/pEtGLsPB6cWZy7) 
image  

- make `get_query_result` (light request with meta data only) for all query results (using `execute_batch`) in order to collect errors, when query selected from list or updated, to show these errors in vs-code editor and to display error icon in tabs [example-2](https://nda.ya.ru/t/ZFeTSKd06cWgxv)
image


packages/ui/src/ui/pages/query-tracker/QueryEditor/QueryEditor.tsx
```diff
@@ -35,12 +38,60 @@ const QueryEditorView = React.memo(function QueryEditorView({
     const activeQuery = useSelector(getQuery);
     const text = useSelector(getQueryText);
     const engine = useSelector(getQueryEngine);
+    const editorErrors = useSelector(getQueryEditorErrors);
+    const decorationsCollection = useRef<monaco.editor.IEditorDecorationsCollection | undefined>(
+        undefined,
+    );
+
+    const updateErrorMarkers = () => {
+        const model = editorRef.current?.getModel();
```
ma-efremoff, line 47: It is better to add model to list of dependencies in line 93.
Also, as I know, `eslint` should highlight some similar problems for useEffect.

packages/ui/src/ui/pages/query-tracker/module/api.ts
```diff
@@ -406,13 +409,43 @@ export function getQueryResultMeta(
...
+    inds: number[],
+): ThunkAction<Promise<QueryResultMeta[]>, RootState, any, any> {
+    return async (_dispatch, getState) => {
+        const state = getState();
+        const {stage} = getQueryTrackerRequestOptions(state);
+        const requests: any[] = inds.map((ind) => ({
+            command: 'get_query_result',
+            parameters: {stage, query_id, result_index: ind},
+        }));
+        const resp = (await ytApiV4Id.executeBatch(YTApiId.getQueryResults, {
```
ma-efremoff, line 430: It looks like the code doesn't handle some errors like 'Network error' (and others)

packages/ui/src/ui/pages/query-tracker/module/query_result/actions.ts
```diff
@@ -242,3 +250,35 @@ export function updateQueryResult(
...
+        if (!result_count) return;
+
+        const inds: number[] = [];
+        for (let ind = 0; ind < result_count; ind++) {
+            if (!hasQueryResult(state, query.id, ind)) inds.push(ind);
+        }
+
+        if (inds.length === 0) return;
+
+        const resultsMeta = await dispatch(getQueryResultMetaList(query.id, inds));
```
ma-efremoff, line 270: What about error handling?

packages/ui/src/ui/pages/query-tracker/QueryEditor/QueryEditor.tsx
shishovKA: I've wrapped `updateErrorMarkers` with `useCallback`, and added it as the dependency in line 93👍🏻

packages/ui/src/ui/pages/query-tracker/module/api.ts
shishovKA: If `executeBatch` fails, this method will return rejected promise. And I will handle error's in other places (where this function is used).

Also I decided to return full `splitedBatchResults` (not only `results` field)

packages/ui/src/ui/pages/query-tracker/module/query_result/actions.ts
shishovKA: I've wrapped  call of `getQueryResultMetaList` with `wrapApiPromiseByToaster`, to show Network Errors etc in toast.

Also added splited error handling for failed sub-requests.

packages/ui/src/ui/pages/query-tracker/QueryEditor/QueryEditor.tsx
ma-efremoff: It is a good practice to add all closured variables to dependencies, at the moment model/editorRef is not added. Does `eslint` hightlight some warning about it?
What is the reason to have `updateErrorMarkers` as a named variable (the function might be used as first argument of useEffect in line 93)?

packages/ui/src/ui/pages/query-tracker/module/query_result/actions.ts
```diff
@@ -242,3 +250,56 @@ export function updateQueryResult(
...
+                dispatch(getQueryResultMetaList(query.id, inds)),
+                {
+                    toasterName: `load_query_results`,
+                    skipSuccessToast: true,
+                    errorTitle: `Failed to load query results meta`,
+                },
+            );
+            forEach_(results, (meta) => {
+                if (meta.error) {
+                    dispatch({
```
ma-efremoff, line 281: Each dispatch call enforces to pass an action through all reducers of redux-store, also some react-components might be rerendered for each such call. Probably it is better to collect errors and use one dispatch-call in such cases.

packages/ui/src/ui/pages/query-tracker/module/query_result/actions.ts
```diff
@@ -242,3 +250,56 @@ export function updateQueryResult(
...
+                            queryId: meta.id,
+                            index: meta.result_index,
+                            error: meta.error as Error,
+                        },
+                    });
+                }
+            });
+            if (error) {
+                forEach_(error.inner_errors, (e, ind) => {
+                    dispatch({
```
ma-efremoff, line 293: The same case as above

packages/ui/src/ui/pages/query-tracker/module/api.ts
```diff
@@ -406,13 +409,41 @@ export function getQueryResultMeta(
...
             setup: getQTApiSetup(),
         });
     };
 }
 
+export function getQueryResultMetaList(
+    query_id: string,
+    inds: number[],
+): ThunkAction<Promise<SplitedBatchResults<QueryResultMeta>>, RootState, any, any> {
+    return async (_dispatch, getState) => {
```
ma-efremoff, line 423: If `_dispatch` is not use it might be just a function returning a promise (not ThunkAction)

packages/ui/src/ui/pages/query-tracker/module/query_result/actions.ts
shishovKA: Have reworked this part
Now I combine all errors in record, and then call dispatch one time.
Also added new reducer type:  SET_QUERY_RESULTS_ERRORS
for these purpose

packages/ui/src/ui/pages/query-tracker/module/query_result/actions.ts
```diff
@@ -242,3 +257,55 @@ export function updateQueryResult(
...
+                    resultsErrors[meta.result_index] = {
+                        error: meta.error as Error,
+                        state: QueryResultState.Error,
+                        resultReady: false,
+                    };
+                }
+            });
+            if (error) {
+                forEach_(error.inner_errors, (e, ind) => {
+                    resultsErrors[inds[errorIndices[ind]]] = {
```
ma-efremoff, line 298: `resultsErrors[inds[errorIndices[ind]]]` 
It is hard to read code-sentences like this, also it looks like a probable source of unwanted exceptions. The code would be more clear and compact if `getQueryResultMetaList` returned batch result "as is" without handling by `splitBatchResult`.

packages/ui/src/ui/pages/query-tracker/module/api.ts
```diff
@@ -406,13 +409,41 @@ export function getQueryResultMeta(
...
+        const state = getState();
+        const {stage} = getQueryTrackerRequestOptions(state);
+        const requests: any[] = inds.map((ind) => ({
```
ma-efremoff, line 426: It is always bad idea to use `any` for type-declarations of any variable, please describe the command in `BatchSubRequest` instead of it.

ma-efremoff (approved):

PR #134, opened 2023-09-04
some fixes

ma-efremoff:
https://nda.ya.ru/t/8V1VzjmU6f9DzH


packages/ui/src/ui/pages/system/NodeQuad/NodeQuad.tsx
```diff
@@ -17,6 +17,7 @@ interface Props {
         | 'mixed'
         | 'registered'
         | 'unregistered'
+        | 'unvoting'
```
shishovKA, line 20: Is it correct theme value?


[here](https://github.com/ytsaurus/ytsaurus-ui/pull/134/commits/4d69b22b75fdfc9ae2f632489cbf23da5b561ca0#diff-8112bb582b99aeefe0a4f439f5fcf4aa9d02178a73480ef05239e7fc1a045107R54)
 i see usage as `nonvoting`

shishovKA:
`[fix(Navigation): add ability to remove table from current path [YTFRONT-3825]`

this commit has wrong ticket number (also in Arcadia)

It looks like [YTFRONT-3837]

packages/ui/src/ui/pages/components/tabs/node/NodeMeta/NodeMeta.tsx
```diff
@@ -69,7 +69,7 @@ function NodeMeta({
     const stateText = hammer.format['FirstUppercase'](state);
     const stateTheme = getStateTheme(state);
 
-    const alertsCount = alerts?.length || 0;
+    //const alertCount = alerts?.length || 0;
```
shishovKA, line 72: should be removed?

shishovKA (approved):

PR #138, opened 2023-09-12
Odin returns

shishovKA:


packages/ui/src/server/controllers/availability.ts
```diff
@@ -0,0 +1,66 @@
...
+            );
+        })
+        .catch((e) => {
+            req.ctx.logError(`Error of getting ${url}`, e);
+            return [];
+        });
+}
+
+export async function getClustersAvailability(req: Request, res: Response) {
+    const odinPath = getApp().config?.odinBaseUrl;
```
ma-efremoff, line 56: All controllers may access config through `req.ctx.config`. It is better to avoid using of `getApp().config` when it is possible. `getApp` is a hack that was brought to minimize code-changes for typescript and break circular dependencies.

ma-efremoff (approved):

PR #143, opened 2023-09-14
fix(Components/Versions): fix for cluster_node/node

ma-efremoff:
https://nda.ya.ru/t/a5wDaCuj6hBXiF


shishovKA:
It would be nice to correct style in versionsID column like this: (center flex container, and remove strange -2pc margin)

image

shishovKA (commented):
I checked Ada cluster, and it looks like 'Nodes' column disappeared
Does it has no `cluster_node` or `node` data?

shishovKA (approved):

PR #168, opened 2023-10-20
fix: allow column groups create form only for `map_nodes` [YTFRONT-3901]

shishovKA:
https://nda.ya.ru/t/u1L3GqEV6qsUDY

### DIFF
--- a/packages/ui/src/ui/components/ACL/ACL.js
+++ b/packages/ui/src/ui/components/ACL/ACL.js
@@ -78,6 +78,7 @@ class ACL extends Component {
 
         path: PropTypes.string.isRequired,
         idmKind: PropTypes.string.isRequired,
+        nodeType: PropTypes.string.isRequired,
         version: PropTypes.string,
         disableAclInheritance: PropTypes.oneOfType([PropTypes.object, PropTypes.bool]),
         bossApproval: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


packages/ui/src/ui/components/ACL/ACL.js
```diff
@@ -429,13 +430,16 @@ class ACL extends Component {
...
+        const {columnGroups, idmKind, path, loadAclData, cluster, aclRequestOptions, nodeType} =
+            this.props;
         const {useEffective} = aclRequestOptions;
+        const allowedNodeTypes = ['map_node'];
         const props = {
             path,
             loadAclDataFn: () => loadAclData({path, idmKind}),
             columnGroups,
             cluster,
+            hasCreateForm: allowedNodeTypes.includes(nodeType),
```
ma-efremoff, line 442: At the moment `allowedNodeTypes`  looks like an unnecessary variable:
`nodeType === 'map_node'`

packages/ui/src/ui/components/ACL/ColumnGroups/ColumnGroups.tsx
```diff
@@ -25,9 +25,16 @@ interface Props {
...
+    hasCreateForm?: boolean;
 }
 
-export default function ColumnGroups({columnGroups, path, loadAclDataFn, cluster}: Props) {
+export default function ColumnGroups({
+    columnGroups,
+    path,
+    loadAclDataFn,
+    cluster,
+    hasCreateForm = false,
```
ma-efremoff, line 36: It looks like editing is still allowed for all node types.
Will be an edited item applied only for current path?
Do we need to allow edit only for `map_nodes`?

shishovKA: by now, we are expecting to have column groups only for 'map_nodes'

i've asked the team whether should i remove the whole Column groups component for other node types, but they asked not to do it right now. They have some plans for it (maybe in some future, they will provide smth like @effective_column_groups attribute to display it for tables and other child nodes)

So i think, we are safe not to worry about editing, because user will be not allowed to create column groups for that nodes -> so the table will be empty

shishovKA: I changed my mind)

I am going to hide actions column (delete and edit) for other node types
Also will rename `hasCreateForm` prop to `allowEditing` or smth more suitable to handle both: hiding Add+ button and action buttons

ma-efremoff (approved):

PR #184, opened 2023-11-07
UIFactory.getAllowedExperimentalPages [YTFRONT-3912]

ma-efremoff:


packages/ui/src/ui/store/selectors/global/index.ts
```diff
@@ -113,15 +117,23 @@ export const isDeveloperSettings = createSelector(
...
-export const getAdminPages = () => UIFactory.getAdminPages();
-
 export const isDeveloper = createSelector(
     [isDeveloperSettings, getSettingsRegularUserUI],
     (isDeveloper, regularUserUI) => {
         return !regularUserUI && isDeveloper;
     },
 );
 
+export const isQueryTrackerAllowed = createSelector(
```
shishovKA, line 127: in this method i would recommend some naming improvements:
- `adminPages` -> `expPages`
- `allowPages` -> `allowedPages`

Also I am worried about part of condition: `!adminPages.includes(Page.QUERIES)`

Why QT page should be allowed if it is not in `adminPages(expPages)` list?

shishovKA: I would suggest condition like this:
`expPages.includes(Page.QUERIES) &&  (isDeveloper || allowPages.includes(Page.QUERIES))`

ma-efremoff: It is better in first check the easiest conditions `isDeveloper` is the one.

ma-efremoff:
`git rebase origin/main`

shishovKA (approved):

PR #191, opened 2023-11-13
some fixes

ma-efremoff:
https://nda.ya.ru/t/WIHjYJv56x6FwJ


packages/ui/src/ui/pages/operations/OperationDetail/tabs/Jobs/OperationJobsTable/OperationJobsTable.scss
```diff
@@ -16,7 +16,6 @@
                 &_type {
```
shishovKA, line 16: probably should be camel case
https://nda.ya.ru/t/_Q7yasgJ6xbXD9

shishovKA (approved):

packages/ui/src/ui/pages/operations/OperationDetail/tabs/Jobs/OperationJobsTable/OperationJobsTable.scss
ma-efremoff: At the moment we use `hammer.format.ReadableField`, it is common approach to format string values. And I think it is better to use it here too.

PR #194, opened 2023-11-14
Do not use pooling for inactive tabs of browser [YTFRONT-3835]

ma-efremoff:


shishovKA (approved):
looks good for me
also e2e test are green

PR #217, opened 2023-12-01
feat: implement an OAuth authorize

draedful:
https://nda.ya.ru/t/R668FkL_74KJ3V

### DIFF
--- a/packages/ui/src/server/utils/authorization.ts
+++ b/packages/ui/src/server/utils/authorization.ts
@@ -1,0 +1,32 @@
+import type {Request} from 'express';
+import {AppMiddleware} from '@gravity-ui/expresskit';
+import {AuthWay} from '../../shared/constants';
+import {isUserOAuthLogged} from '../components/oauth';
+
+export function isAuthorized(req: Request) {
+    if (req.yt) {
+        return Boolean(Object.keys(req.yt.ytApiAuthHeaders ?? {}));
+    }
+    return false;
+}
+
+export function getAuthWay(req: Request): AuthWay | undefined {
+    if (!isAuthorized(req)) {
+        return undefined;
+    }
+    if (isUserOAuthLogged(req)) {
+        return 'oauth';
+    }
+    return 'passwd';
+}
+
+export function authorizationResolver(resolver: AppMiddleware): AppMiddleware {
+    return (req, res, next) => {
+        if (isAuthorized(req)) {
+            next();
+            return;
+        }
+        resolver(req, res, next);
+        return;
+    };
+}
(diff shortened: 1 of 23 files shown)


packages/ui/src/server/utils/authorization.ts
```diff
@@ -0,0 +1,32 @@
...
+export function isAuthorized(req: Request) {
+    if (req.yt) {
+        return Boolean(Object.keys(req.yt.ytApiAuthHeaders ?? {}));
+    }
+    return false;
+}
+
+export function getAuthWay(req: Request) {
+    if (!isAuthorized(req)) {
+        return false;
```
ma-efremoff, line 15: `return undefined;`

packages/ui/src/shared/constants/index.ts
```diff
@@ -3,6 +3,14 @@ export const YT_CYPRESS_COOKIE_NAME = 'YTCypressCookie';
 export const YT_LOCAL_CLUSTER_ID = 'ui';
 export const ODIN_PAGE_ID = 'odin';
 
+export const YT_OAUTH_ACCESS_TOKEN_NAME = 'yt_oauth_access_token';
+export const YT_OAUTH_REFRESH_TOKEN_NAME = 'yt_oauth_refresh_token';
+
+export enum AuthWay {
+    OAuth = 'oauth',
+    YT = 'yt',
+}
```
ma-efremoff, line 12: It is always better to use `const enum` or union of strings type instead of regular enums  when it is possible, cause it allows to avoid unnecessary links between modules. `const enum`-s and union of strings [generate no code after compilation](https://www.typescriptlang.org/play?#code/KYOwrgtgBAggQgYSgbwFBQ1AhlAvFAciwIBp1MAjPQi08jAY2oIYNQF9VUGB7EAZwAuUUJCgARAKIAxFPSgATZgrqYRzYKswAzZtradUggJ4AHYFADiACQCSzAOYEoAH0IALZ24IBLAgG4uXgFhLAAuWERqeAQAOixA4KFFCKlZfDTYhUS+ZIdHANQgA). 

Also it is better to use more specific value instead of `'yt'` like `'yt-passwd'`/`'yt-pwd'` it allows easier to find places of usage in the code like [authWay === 'yt'](https://github.com/ytsaurus/ytsaurus-ui/pull/217/files#diff-60df5594b756d491317c43eca0b4e3e368e0728b15b706a9c2328db977c288c1R164). 
By the way you use comparison with a literal string so union of strings is more natural:
```ts
type AuthWay = 'YT_OAUTH' | 'YT_PASSWD';
```

ma-efremoff (approved):

Issue #208, opened 2023-11-29
Query results with more than 50 columns are not shown properly


zlobober:
Create a table with 100 columns using the script below:
```
#!/usr/bin/python3
import yt.wrapper as yt
n = 100
schema = [
    {"name": "a" + str(i), "type": "int64"} for i in range(n)
]
yt.create("table", "//home/max/many_columns", attributes={"schema": schema}, force=True)
yt.write_table(
    "//home/max/many_columns",
    [{"a" + str(i): i for i in range(n)}]
)
```

Run the following query (in YQL or CHYT dialect):
```
select * from `//home/max/many_columns`
```

Only first 50 columns are shown:
![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/a01fe86b-6d70-41d5-b0c9-d9d071de91fc)

Opening this table in Navigation leads to the correct result:

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/27e93f32-4dc4-4737-91b7-5c46f2c8c99e)

I believe that the difference is somewhere around the options of web_json when requesting the result. Comparison of parameters of `read_query_result` request on Queries page and `read_table` on Navigation should unveil the difference.

PR #220, opened 2023-12-05
Query results with more than 50 columns are not shown properly #208

vitshev:
According to issue #208, this PR addresses the difference between two queries: read_query_result and read_table.

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/module/api.ts
+++ b/packages/ui/src/ui/pages/query-tracker/module/api.ts
@@ -301,6 +301,7 @@ export function readQueryResults(
                         value_format: 'yql',
                         field_weight_limit: settings?.cellsSize,
                         encode_utf8: 'false',
+                        max_selected_column_count: 3000,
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/pages/query-tracker/module/api.ts
```diff
@@ -301,6 +301,7 @@ export function readQueryResults(
                         value_format: 'yql',
                         field_weight_limit: settings?.cellsSize,
                         encode_utf8: 'false',
+                        max_selected_column_count: 3000,
```
vitshev, line 304: https://github.com/ytsaurus/ytsaurus-ui/blob/07def6174c9a5a9356b39e638e2ad23cce7c8e6b/packages/ui/src/ui/utils/navigation/content/table/table.js#L71

vitshev:
@ma-efremoff @yandex-moonbreeze

ma-efremoff (approved):

PR #231, opened 2023-12-08
Add odin base url passing to common.ts config

mutexed:
Added passing odinBaseUrl using ODIN_BASE_URL env.

### DIFF
--- a/packages/ui/src/server/configs/common.ts
+++ b/packages/ui/src/server/configs/common.ts
@@ -67,6 +67,8 @@ const config: Partial<AppConfig> = {
         },
     },
 
+    odinBaseUrl: process.env.ODIN_BASE_URL,
+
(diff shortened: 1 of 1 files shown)


ma-efremoff (commented):
Our CHANGELOG.md is autogenerated by titles of commits, we use [`Conventional Commits`](https://www.conventionalcommits.org/en/v1.0.0/). Please replace the title of your commit:
`done` -> `feat(server/configs): use process.env.ODIN_BASE_URL to set odinBaseUrl`

zlobober:
I'd suggest not duplicating the option to env, and rather set it via common.js. Example:
```
common.js: 
module.exports = {"odinBaseUrl":"http://...", ...}
```

Our operator is also able to set it directly from `ui` section of ytsaurus spec.

Issue #226, opened 2023-12-06
Prevent user from closing the tab/browser window with unsaved query text


zlobober:
Subj, must be done by native browser "there may be unsaved changes" window.

UX must be following:
* if you opened a new clean window, closing is ok
* if you started typing something and the text is non-empty, closing is not ok
* if you pressed "run" or "save draft" buttons, the state is clean again, so closing is ok
* if you edited text further from runned or saved state, the state is dirty, so closing is not ok
* if you opened a query by a link, it is clean, so closing is ok

PR #252, opened 2023-12-18
feat: Prevent user from closing the tab/browser window with unsaved query text #226

vitshev:
#226

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/module/query/reducer.ts
+++ b/packages/ui/src/ui/pages/query-tracker/module/query/reducer.ts
@@ -26,6 +26,7 @@ export interface QueryState {
         cluster?: string;
         useDraft?: boolean;
     };
+    dirtySinceLastSubmit: boolean;
     state: 'init' | 'loading' | 'ready' | 'error';
 }
 
@@ -40,6 +41,7 @@ const initState: QueryState = {
     queryItem: undefined,
     draft: {...initialQueryDraftState},
     params: {},
+    dirtySinceLastSubmit: false,
     state: 'init',
 };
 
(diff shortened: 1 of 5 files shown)


packages/ui/src/ui/pages/query-tracker/module/query/reducer.ts
```diff
@@ -89,12 +92,15 @@ export function reducer(state = initState, action: Actions): QueryState {
             };
         }
         case SET_QUERY_PATCH: {
```
vitshev, line 94: It seems to me that those two actions SET_QUERY_PATCH/SET_QUERY cover all of our cases.

packages/ui/src/ui/hooks/use-before-unload.ts
```diff
@@ -0,0 +1,18 @@
+import {useEffect} from 'react';
+
+const beforeUnloadHandler = (e: BeforeUnloadEvent) => {
+    e.preventDefault();
+
+    // Included for legacy support, e.g. Chrome/Edge < 119
+    e.returnValue = true;
+};
+
+export const useBeforeUnload = ({shouldListen}: {shouldListen: boolean}) => {
```
ma-efremoff, line 10: Probably it is better to name the hook `usePreventUnload`

vitshev: renamed it

packages/ui/src/ui/pages/query-tracker/module/query/reducer.ts
ma-efremoff: All changed parts of the state that received from api should prevent unload. I nearly sure that some data for SET_QUERY_PARAMS comes from api. Could you please check what parts of the state comes from api?

vitshev: It seems that SET_QUERY_PARAMS only works with the GET parameters of the page.

ma-efremoff (approved):

ma-efremoff (approved):

Issue #204, opened 2023-11-29
Display location ids on data node sidebar


zlobober:
Page: /components/nodes?nodeType=data_nodes
The sidebar contains various information about node locations, but does not contain location uuids.
![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/d5e39469-29e7-447e-ac11-a24c0dee7508)

Id may be taken from @chunk_locations attribute of the data node.

PR #260, opened 2023-12-19
feat: display location ids on data node table [#204]

vitshev:
https://github.com/ytsaurus/ytsaurus-ui/issues/204


vitshev:
@ma-efremoff Could you tell me please. I've noticed that the [original issue](https://github.com/ytsaurus/ytsaurus-ui/issues/204) contains an outdated screenshot of the interface, and I actually need to add location ids to the table?

Issue #229, opened 2023-12-06
Lack of proper rendering for complex types in Navigaion


zlobober:
OSS version lacks the possibility to properly display complex types in navigation. At the same time, complex values are displayed correctly through the Queries page.

Repro:
```
$ yt create table "//home/max/test_complex_types" --attributes '{schema=[{name=d; type_v3=date}]}' 
7b7-3c651-13440191-d0980a17
$ yt write-table "//home/max/test_complex_types" --format yson
{d=19697}
```

Opening the table in the navigation renders complex type as an underlying int value:

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/3679a099-0aff-4f03-9642-c95c6454a9b9)

Querying the table via CHYT/YQL works well:

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/4c78b533-4b87-41ce-aa5d-f2a94ba91a8a)

It looks like the internal option "Use YQL v3 types" is missing, even though the support of proper rendering is available in OSS version. Type V3 rendering should be on by default, and there should be an option to disable the type v3 rendering, i.e. this feature must be opt-out.

PR #265, opened 2023-12-20
fix: render complex types in navigation #229

vitshev:


packages/ui/src/ui/store/reducers/global/cluster-ui-config.ts
```diff
@@ -7,4 +7,6 @@ export interface ClusterUiConfig {
     enable_maintenance_api_proxies?: boolean;
 }
 
-export const defaultClusterUiConfig: Partial<ClusterUiConfig> = {};
+export const defaultClusterUiConfig: Partial<ClusterUiConfig> = {
+    web_json_value_format: 'yql',
```
vitshev, line 11: @ma-efremoff I want to ask, is it normal to enable this setting for everyone?

ma-efremoff: The option was required when not all versions of running clusters supported the format. It might be removed 
 if we sure all clusters are are ready for that.

ma-efremoff: It looks like we can to remove the option.

packages/ui/src/ui/containers/SettingsPanel/settings-description.tsx
```diff
@@ -395,16 +393,6 @@ function useSettings(cluster: string, isAdmin: boolean): Array<SettingsPage> {
...
-                isSupportedYqlTypes &&
-                    makeItem(
-                        'YQL V3 types',
-                        'top',
-                        <SettingsMenuItem
-                            settingName={SettingName.DEVELOPMENT.YQL_TYPES}
-                            settingNS={NAMESPACES.DEVELOPMENT}
-                            oneLine
-                        />,
-                    ),
```
ma-efremoff, line 407: The option is still used from `isYqlTypesEnabled` selector.

vitshev: oh, I got it. Should I just remove the isSupportedYqlTypes?

ma-efremoff: yes, only `isSupportedYqlTypes` should be removed

vitshev: Returned it

ma-efremoff (approved):

PR #272, opened 2023-12-23
fix: btoa method for NodeJS environment

shishovKA:
### Problem:
When _yt-javascript-wrapper_ request query contains some cyrillic symbols:
`* from [//home/someTable] where x = "привет"`
the result of this request might be incorrect in NodeJs projects, but it works fine with browser environment.
_yt-javascript-wrapper_ library has redefinition of `btoa` method which takes effect only for NodeJS projects.
Redefined method works differently as `window.btoa`

### DIFF
--- a/packages/javascript-wrapper/lib/utils/setup.js
+++ b/packages/javascript-wrapper/lib/utils/setup.js
@@ -131,24 +131,19 @@ setup.createOption('requestHeaders', 'object', {});
 
 setup.createOption('suppressAccessTracking', 'boolean', false);
 
-var btoa = function (string) {
-    return Buffer(string).toString('base64');
-};
-
-if (typeof window !== 'undefined') {
-    btoa = window.btoa;
-}
-
 setup.encodeForYt = function (str) {
-    return unescape(encodeURIComponent(str));
+    if (typeof window !== 'undefined') {
+        // https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa#Unicode_strings
+        return window.btoa(unescape(encodeURIComponent(str)));
+    }
+    return Buffer.from(str).toString('base64');
 };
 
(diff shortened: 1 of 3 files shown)


packages/javascript-wrapper/lib/utils/setup.js
```diff
@@ -132,7 +132,7 @@ setup.createOption('requestHeaders', 'object', {});
 setup.createOption('suppressAccessTracking', 'boolean', false);
 
 var btoa = function (string) {
-    return Buffer(string).toString('base64');
+    return Buffer.from(string, 'binary').toString('base64');
```
ma-efremoff, line 135: According to [the documentation](https://nodejs.github.io/nodejs.dev/en/api/v18/buffer/#buffers-and-character-encodings) `binary` is legacy alias for `latin1`. And `utf8` is default value, so it is better to pass `utf8` as second parameter or omit it.

shishovKA: We cannot use `Buffer.from(string, 'utf8')` here, because in combination with `unescape(encodeURIComponent(...))` it will produce different result.

Although `Buffer(string, 'binary').toString('base64')` works like `window.btoa(...)`, i agree not to use it due to its legacy 👍🏻.

So in second iteration i suggest some refactoring of `setup.encodeForYt` method, where:
 - `window.btoa(unescape(encodeURIComponent(str)))` will be used for browser environment (as well known [browser hack](https://ecmanaut.blogspot.com/2006/07/encoding-decoding-utf8-in-javascript.html) to encode utf-8)
 - and `Buffer.from(string).toString('base64')` for Node js utf-8 encoding

ma-efremoff:
It looks like the fix breaks the UI. Have you tested it with yt-local cluster (`npm run dev:localmode`)?

PR #288, opened 2024-01-17
fix: add bundle resources validation

vansergen:
https://nda.ya.ru/t/FlGlj43Q74MbU6

### DIFF
--- a/packages/ui/src/ui/store/actions/tablet_cell_bundles/index.ts
+++ b/packages/ui/src/ui/store/actions/tablet_cell_bundles/index.ts
@@ -145,7 +145,7 @@ export function fetchTabletsBundles(): TabletsBundlesThunkAction {
                     .get(YTApiId.bundleControllerZones, {
                         parameters: {
                             path: `//sys/bundle_controller/controller/zones`,
-                            attributes: ['tablet_node_sizes', 'rpc_proxy_sizes'],
+                            attributes: ['tablet_node_sizes', 'rpc_proxy_sizes', 'data_centers'],
                         },
                         cancellation: cancelHelper.saveCancelToken,
                     })
--- a/packages/ui/src/ui/utils/tablet_cell_bundles/bundles/bundle-editor-dialog/index.ts
+++ b/packages/ui/src/ui/utils/tablet_cell_bundles/bundles/bundle-editor-dialog/index.ts
@@ -4,6 +4,7 @@ import hammer from '../../../../common/hammer';
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import {
     BundleControllerConfig,
+    BundleDataCenter,
     BundleDefaultConfigData,
 } from '../../../../store/reducers/tablet_cell_bundles';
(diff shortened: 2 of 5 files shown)


packages/ui/src/ui/utils/tablet_cell_bundles/bundles/bundle-editor-dialog/index.ts
```diff
@@ -269,6 +270,25 @@ export const simpleBundleValidate = (v?: number | string) => {
...
+    return (input?: number | string): string | undefined => {
+        const number = Number(input);
+        let error: string | undefined;
+
+        if (!Number.isFinite(number)) {
+            error = 'Incorrect value';
+        } else if (number < 0) {
+            error = 'Must be greater than or equal to zero';
+        } else if (length > 0 && number % length !== 0) {
+            error = `Must divide the number of data centers (${length})`;
```
evtaranov, line 285: suggestion: Looks like this is not worded correctly. I would recommend changing for:
`Must be a multiple of the number of data centers (${length})`

packages/ui/src/ui/store/actions/tablet_cell_bundles/index.ts
```diff
@@ -145,7 +145,7 @@ export function fetchTabletsBundles(): TabletsBundlesThunkAction {
                     .get(YTApiId.bundleControllerZones, {
                         parameters: {
                             path: `//sys/bundle_controller/controller/zones`,
-                            attributes: ['tablet_node_sizes', 'rpc_proxy_sizes'],
+                            attributes: ['tablet_node_sizes', 'rpc_proxy_sizes', 'data_centers'],
```
ma-efremoff, line 148: You have to use  `//sys/bundle_controller/controller/zones/zone_default` path instead of `//sys/bundle_controller/controller/zones`. So it is better to use `executeBatch` command here. Also we don't need full contend of the attribute it is enough to use `list` subrequest, like:
```bash
$ yt --proxy cluster_id list //sys/bundle_controller/controller/zones/zone_default/@data_centers
dc1
dc1
dc3
```

packages/ui/src/ui/utils/tablet_cell_bundles/bundles/bundle-editor-dialog/index.ts
```diff
@@ -269,6 +270,25 @@ export const simpleBundleValidate = (v?: number | string) => {
     return error;
 };
 
+export const dcBundleValidate = ({
```
ma-efremoff, line 273: I think it is better to name it like `instanceCountValidate`/`validateInstanceCount`/...

vansergen: f2cf7f75c35c653d9d499b488f7df424006bff8f

packages/ui/src/ui/store/actions/tablet_cell_bundles/index.ts
ma-efremoff: I was wrong, the code works fine.

ma-efremoff (approved):

PR #291, opened 2024-01-18
fix: nodejs error during logout

vitshev:
I get an error during logout

```
[11:31:26 UTC] ERROR: [Express POST] [handleLogout] Error [61626bd4-50e3-4a00-9bac-883324230dec]
    extra: {}
    err: {
      "type": "Object",
      "message": "Cannot set headers after they are sent to the client",
      "stack":
          Error: Cannot set headers after they are sent to the client
              at new NodeError (node:internal/errors:405:5)
              at ServerResponse.setHeader (node:_http_outgoing:648:11)
```

According to the documentation https://expressjs.com/ru/api.html#res.sendStatus the response status is sent along with the body


vansergen (approved):

ma-efremoff (approved):

vansergen:
Closes #292

Issue #294, opened 2024-01-19
if show in ui long yson, the lines will moving out


vitshev:
if UI shows in long yson (string in other rows longer than first screen), the lines will move out

```
$node = Json(@@
  {
    "abc": {"def": 123, "ghi": "привет"},
    "abc1": {"def": 123, "ghi": "привет1"},
    "abc2": {"def": 123, "ghi": "привет12"},
    "abc3": {"def": 123, "ghi": "привет123"},
    "abc4": {"def": 123, "ghi": "привет1234"},
    "abc5": {"def": 123, "ghi": "привет12345"},
    "abc6": {"def": 123, "ghi": "привет123456"},
    "abc7": {"def": 123, "ghi": "привет1234567"},
    "abc8": {"def": 123, "ghi": "привет12345678"},
    "abc9": {"def": 123, "ghi": "привет123456789"},
    "abc10": {"def": 123, "ghi": "привет1234567891"},
    "abc11": {"def": 123, "ghi": "привет12345678912"},
    "abc12": {"def": 123, "ghi": "привет123456789123"},
    "abc13": {"def": 123, "ghi": "привет1234567891234"},
    "abc14": {"def": 123, "ghi": "привет12345678912345"},
    "abc15": {"def": 123, "ghi": "привет123456789123456"}
  }
@@);
select $node, 1 as a, 'b' as bb
```

<img width="506" alt="0eb298ee-0aed-4aba-b695-6db4bf295b4c" src="https://github.com/ytsaurus/ytsaurus-ui/assets/3217636/58c87ba7-8973-4a3b-9e13-72e4218d99e9">
<img width="506" alt="6aabe731-e30f-4b07-ba92-cb233725c0fd" src="https://github.com/ytsaurus/ytsaurus-ui/assets/3217636/9c779f38-2d66-4e0b-9ccc-2ab674e96743">

PR #295, opened 2024-01-19
fix: synchronize table header when resizing [#294]

vitshev:
https://github.com/ytsaurus/ytsaurus-ui/issues/294

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/QueryResultsView/YQLTable/YQLTable.tsx
+++ b/packages/ui/src/ui/pages/query-tracker/QueryResultsView/YQLTable/YQLTable.tsx
@@ -31,6 +31,10 @@ function rowKey(row: unknown, index: number) {
     return index;
 }
 
+function triggerWindowResize() {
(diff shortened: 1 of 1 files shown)


ma-efremoff:
It is always better to provide some scope in titles of commits, like:
`fix(Queries/Results): synchronize table header when resizing [#294]`

packages/ui/src/ui/pages/query-tracker/QueryResultsView/YQLTable/YQLTable.tsx
```diff
@@ -379,6 +386,10 @@ function ShowMoreInline({formattedValue, strippedDown}: ShowMoreInlineProps) {
             <Link
                 onClick={() => {
                     setShowFull((v) => !v);
+
+                    if (onClick) {
+                        onClick();
```
ma-efremoff, line 391: `onClick?.();`

ma-efremoff (approved):

PR #328, opened 2024-02-01
[FIX]: Use the default value when `clock_cell` is missing

vansergen:
https://nda.ya.ru/t/hV1gQeSN74WAX9


packages/ui/src/ui/store/actions/system/masters.ts
```diff
@@ -349,7 +349,10 @@ export function loadMasters() {
     };
 }
 
-function getCellIdTag(uuid: string): number {
+function getCellIdTag(uuid?: string): number {
+    if (!uuid) {
+        return 0;
```
ma-efremoff, line 354: It is better to return `undefined`, and show it in UI as `format.NO_VALUE`

ma-efremoff (approved):

PR #367, opened 2024-02-19
chore: update README

vitshev:


packages/ui/README.md
```diff
@@ -10,7 +10,9 @@ Additionally you have to provide `secrets/yt-interface-secret.json` file with [a
 
 ```json
 {
-  "oauthToken": "special-user-secret-token"
```
ma-efremoff, line 13: The field is still [actual](https://github.com/ytsaurus/ytsaurus-ui/blob/main/packages/ui/src/server/components/requestsSetup.ts#L18), cluster-specific tokens are optional.

```ts
 require(ytInterfaceSecret)?.oauthToken; // Backward compatibility
```
It is better to get rid of the comment. It is not backward compatibility, it is just another approach that allows to have one common token for all clusters.

packages/ui/README.md
```diff
@@ -10,7 +10,9 @@ Additionally you have to provide `secrets/yt-interface-secret.json` file with [a
...
+    "CLUSTER_ID": {
```
ma-efremoff, line 13:
```json
{
    // comon ooauth token
    "oauthToken": "special-user-secret-tocken",
    "cluster_id1": {
        // cluster_id1 specific oauth token
        "oauthToken": "cluster1-special-user-secret-token",
    },
    "cluster_id2": {
        // cluster_id2 specific oauth token
        "oauthToken": "cluster2-special-user-secret-token",
    },
    ...
}
```

PR #373, opened 2024-02-22
feat(Navigation): added the ability to edit a document [YTFRONT-3921]

SimbiozizV:


packages/ui/src/ui/pages/navigation/content/Document/DocumentEditModal.tsx
```diff
@@ -0,0 +1,61 @@
...
+        result.current = data.value || '';
+        setError(Boolean(data.error));
+    };
+
+    const handleSaveDocument = () => {
+        onSave(result.current);
+    };
+
+    return (
+        <Dialog onClose={onCancel} open={open} modalClassName={block()}>
```
ma-efremoff, line 34: We have special component to generate forms, it is better to use it here:
```ts
const [visible, setVisible] = React.useState(false);
...
return (
     <YTDFDIalog
          visible={visible}
          onClose={() => setVisible(false)}
          onAdd={() => {...}}
          fields=[{name: 'value', type: 'json', caption: 'Caption'}]
     />
```
And we don't need to remember `editing`/`saving` state in the redux store, so the following actions are unnecessary: `SET_DOCUMENT_SAVING`, `SET_DOCUMENT_EDIT_MODE`.

ma-efremoff (commented):
All refactoring changes like rewriting code from js to ts should be in a separate commits. And we have to use `git mv` for such files else `git log` history will be interrupted.

ma-efremoff (approved):

PR #396, opened 2024-03-01
some fixes

ma-efremoff:
https://nda.ya.ru/t/sZGW154874peQT

### DIFF
--- a/packages/ui/scripts/sync-peer-dependencies.js
+++ b/packages/ui/scripts/sync-peer-dependencies.js
@@ -1,0 +1,54 @@
+const process = require('node:process');
+const fs = require('node:path');
+
+const [_x, script, action, file] = process.argv;
+
+if (action !== 'check' && action !== 'fix') {
+    console.error(`Unexpecte action: ${action}`);
+    showHelpAndExit(1);
+}
+
+const {PWD} = process.env;
+
+function showHelpAndExit(exitCode) {
+    console.log('\nUsage:');
+    console.log('\t', process.argv0, fs.relative(PWD, script), 'check', file);
+    showFixHelp(exitCode);
+}
+
+function showFixHelp(exitCode) {
+    console.log('\t', process.argv0, fs.relative(PWD, script), 'fix', file);
+    console.log('');
+    process.exit(exitCode);
+}
+
+const packageJson = require(file.startsWith('/') ? file : PWD + '/' + file);
+
+const {dependencies: deps, devDependencies: dev, peerDependencies: peer} = packageJson;
+let needFix = false;
+let fixed = false;
+Object.keys(peer).forEach((key) => {
+    const v = deps[key] ?? dev[key];
+    if (key.startsWith('@') && peer[key] !== v) {
+        switch (action) {
+            case 'check': {
+                needFix = true;
+                console.error('Peer dependency', key, 'is not synced:', v, '!=', peer[key]);
+                break;
+            }
+            case 'fix': {
+                peer[key] = v;
+                fixed = true;
+            }
+        }
+    }
+});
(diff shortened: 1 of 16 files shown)


packages/ui/scripts/sync-peer-dependencies.js
```diff
@@ -0,0 +1,54 @@
...
+    showFixHelp(exitCode);
+}
+
+function showFixHelp(exitCode) {
+    console.log('\t', process.argv0, fs.relative(PWD, script), 'fix', file);
+    console.log('');
+    process.exit(exitCode);
+}
+
+const packageJson = require(file.startsWith('/') ? file : PWD + '/' + file);
```
SimbiozizV, line 25: I would add a try/catch block here since "require" may not find the file you need.

`try {
  const filePath = file.startsWith('/') ? file : `${PWD}/${file}`;
  const packageJson = require(filePath);
  // .......
} catch (error) {
  console.error(`Error loading file: ${file}`, error);
   // .......
}`

packages/ui/src/ui/pages/tablet_cell_bundles/bundles/BundleEditorDialog/BundleEditorDialog.tsx
```diff
@@ -654,25 +637,58 @@ export function BundleEditorDialog() {
...
+    const res: Pick2<typeof values, 'resources', 'info', string> &
+        Pick2<typeof values, 'memory_limits', 'memory_reset', string> = {
+        resources: {info: ''},
+        memory_limits: {memory_reset: ''},
+    };
+
+    const {resources} = values;
+
+    const currentAccLimit =
+        Object.values(values.memory_limits || {}).reduce((acc: number, v: number | boolean) => {
```
SimbiozizV, line 658: Reduce always returns its value. The construction with || 0 is not needed
I suggest we redo it for this:
`const currentAccLimit = Object.values(values.memory_limits || {})
  .filter((v: number | boolean): v is number => typeof v === 'number')
  .reduce((acc: number, v: number) => acc + v, 0);`

packages/ui/scripts/sync-peer-dependencies.js
ma-efremoff: It is OK to fail in such case, the script should not continue code-execution.

packages/ui/src/ui/pages/tablet_cell_bundles/bundles/BundleEditorDialog/BundleEditorDialog.tsx
ma-efremoff: Additional filtering before `reduce`-call is not the best approach cause it brings 2*N instead of N iterations in worst case, also it brings unnecessary allocation of a new array with intermediate elements.

ma-efremoff: Unnecessary `|| 0` is removed

SimbiozizV (approved):

packages/ui/scripts/sync-peer-dependencies.js
```diff
@@ -0,0 +1,54 @@
...
+const fs = require('node:path');
+
+const [_x, script, action, file] = process.argv;
+
+if (action !== 'check' && action !== 'fix') {
+    console.error(`Unexpecte action: ${action}`);
+    showHelpAndExit(1);
+}
+
+const {PWD} = process.env;
```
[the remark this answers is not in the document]
ma-efremoff: The script runs as postinstall, it will be never called from browser.

Issue #368, opened 2024-02-20
YQL query progress shows wrong stage for a query step


Krock21:
It shows awaiting cache
![image (1)](https://github.com/ytsaurus/ytsaurus-ui/assets/14953722/bd672c87-fcd4-4937-964b-1a11d81d558f)
While it should show running, as on timeline and "progress" field of the response
![image (2)](https://github.com/ytsaurus/ytsaurus-ui/assets/14953722/a9bd922d-6f74-44b8-a487-dbf2ec449764)

PR #398, opened 2024-03-05
fix(QueryTracker): YQL query progress shows wrong stage for a query step [#368]

vitshev:


packages/ui/src/ui/pages/query-tracker/Plan/NodeStages.tsx
```diff
@@ -59,7 +59,9 @@ function prepareData(data: Stages, state?: NodeState, finishedAt?: string) {
     const stages: StageRow[] = [];
     let [currentStage, currentTime] = Object.entries(data[0])[0];
     const startedTime = currentTime;
-    for (let i = 1; i < data.length; i++) {
+    const dataLength = Object.keys(data).length;
+
+    for (let i = 1; i < dataLength; i++) {
```
ma-efremoff, line 64: Are the keys in response might me unordered? If not then it is better to iterate through `Object.keys(data)`:
```ts
Object.keys(data).forEach(() => {
    if (i === "0") {
        return;
    }
   ...
})
```
At the moment it is hard to guess type of `data` looking at the code (object or array).

SimbiozizV: can we change the format of the incoming data?
`
    [...,{name: 'Running', time: '2024-02-19T16:22:55.344038Z'}, ...]
`
the current data format is not the most convenient
Object.keys/entries have bigO equal to O(n) which is much larger than Array.length = O(1)

vrozaev: If you are worried about perfomance and you are sure that there are no missing keys in `data` then you can do the next thing:

`for (let i = 1; data[i]; i++) {`

SimbiozizV (approved):

SimbiozizV (approved):

PR #414, opened 2024-03-12
feat(odin): support odin url per cluster

vitshev:


packages/ui/src/server/controllers/availability.ts
```diff
@@ -53,7 +53,13 @@ function getAvailability(req: Request, clusters: Record<string, {id: string}>, o
 }
 
 export async function getClustersAvailability(req: Request, res: Response) {
-    const odinPath = getApp().config?.odinBaseUrl;
+    const {ytAuthCluster: cluster} = req.params;
+    const odinPath = _.get(
+        getApp().config,
+        `odinBaseUrl.${cluster}`,
+        getApp().config.odinBaseUrl,
+    ) as string;
```
ma-efremoff, line 61: Please replace `getApp().config` with `req.ctx.config`, `getApp` was brought to break some cyclic dependencies in our code, it is not recommended to use it.
Also it is better to avoid usages of `_.get/_.set` cause they always break ts-typings:
```ts
const {odinBaseUrl} =  req.ctx.config;
const odinPath = 'string' === odinBaseUrl ? odinBaseUrl : odinBaseUrl?.[cluster];
```

packages/ui/src/server/controllers/odin-proxy-api.ts
```diff
@@ -19,14 +19,19 @@ export async function odinProxyApi(req: Request, res: Response) {
 }
 
 async function odinProxyApiImpl(req: Request, res: Response) {
-    const odinPath = getApp().config?.odinBaseUrl;
+    const {action, ytAuthCluster: cluster} = req.params;
+    const odinPath = _.get(
```
ma-efremoff, line 23: Please see previous comment

packages/ui/src/server/routes.ts
```diff
@@ -55,7 +55,7 @@ const routes: AppRoutes = {
     'GET /api/yt-proxy/:ytAuthCluster/:command': {handler: ytProxyApi},
 
     'GET /api/odin/proxy/:action/:ytAuthCluster?': {handler: odinProxyApi},
-    'GET /api/odin/clusters/availability': {handler: getClustersAvailability},
+    'GET /api/odin/clusters/availability/:ytAuthCluster?': {handler: getClustersAvailability},
```
ma-efremoff, line 58: Any changes in `routes` require corresponding changes in `src/ui`. It looks like a mistake. Do we really need it?

packages/ui/src/server/controllers/availability.ts
```diff
@@ -53,7 +53,13 @@ function getAvailability(req: Request, clusters: Record<string, {id: string}>, o
 }
 
 export async function getClustersAvailability(req: Request, res: Response) {
-    const odinPath = getApp().config?.odinBaseUrl;
+    const {ytAuthCluster: cluster} = req.params;
+    const odinPath = _.get(
```
ma-efremoff, line 57: As I understand `odinPath` should be calculated for each cluster in `getAvailability`, shouldn't it?

packages/ui/src/server/controllers/availability.ts
```diff
@@ -26,41 +25,81 @@ function getClusterAvailability(clusterConfig: {id: string}, odinPath: string) {
...
             return Promise.all(
                 _.map(clusters, (clusterConfig) => {
-                    const id = clusterConfig.id;
-                    return getClusterAvailability(clusterConfig, odinPath)
-                        .then((availability) => ({id, availability}))
-                        .catch((error) => {
-                            req.ctx.logError('getAvailability error', error);
-                            return {id};
-                        });
+                    if (isMultiOdinBaseUrl) {
```
ma-efremoff, line 56:
```ts
_.map(cluster, (clusterConfig) => {
    const odinPath = isMultiOdinBaseUrl ? odinBaseUrl : odinBaseUrl?.[clusterConfig.id];
    if (!odinPath) {
        return {};
    }
    const alive = isMultiOdinBaseUrl ? axios.get(makeOdinIsAlive(odinPath)) : {data: true};
    return alive.then(({data}) => {
        if (data !== true) {
            return {};
        }
        return getClusterAvailability(clusterConfig, odinPath)
            .then((availability) => ({id, availability}))
            .catch((error) => {
                req.ctx.logError('getAvailability error', error);
                return {id};
            });
    });
})
```

packages/ui/src/server/controllers/availability.ts
```diff
@@ -26,41 +25,69 @@ function getClusterAvailability(clusterConfig: {id: string}, odinPath: string) {
...
+                    return alive.then(({data}) => {
+                        if (data != true) {
+                            return {};
+                        }
+
+                        return getClusterAvailability(clusterConfig, odinPath)
+                            .then((availability) => ({id: clusterConfig.id, availability}))
+                            .catch((error) => {
+                                req.ctx.logError('getAvailability error', error);
+                                return {id: clusterConfig.id};
```
ma-efremoff, line 69:
```ts
return {};
```

ma-efremoff (approved):

PR #426, opened 2024-03-19
fix: request for settings is made without a cluster

vitshev:
Sometimes, the cluster doesn't have enough time to be set, and requests go through without a cluster.

### DIFF
--- a/packages/ui/src/ui/containers/ClusterPage/ClusterPage.js
+++ b/packages/ui/src/ui/containers/ClusterPage/ClusterPage.js
@@ -150,6 +150,9 @@ class ClusterPage extends Component {
         } = this.props;
 
         return Promise.resolve().then(() => {
+            initClusterParams(cluster);
+            updateTitle({cluster});
+
             // todo: get rid of redirectToBetaSwitched setting.
             // It`s exist for set default value of redirectToBeta setting, when settings document is empty yet.
             // Set redirectToBeta on server after get all user settings (home.js). Or get rid of this logic at all.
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/store/selectors/global/index.ts
```diff
@@ -214,6 +214,11 @@ export const getGlobalAsideHeaderWidth = (state: RootState) => state.global.asid
...
 
-export const getSettingsCluster = createSelector([getCluster], (cluster) => {
-    return userSettingsCluster ?? cluster;
-});
+export const getParamsCluster = (state: RootState): string => state?.global?.paramsCluster;
+
+export const getSettingsCluster = createSelector(
+    [getParamsCluster, getCluster],
+    (paramsCluster, cluster) => {
+        return userSettingsCluster || cluster || paramsCluster;
```
ma-efremoff, line 222: Do we really need the change to fix the issue?

packages/ui/src/ui/containers/ClusterPage/ClusterPage.js
```diff
@@ -150,23 +150,28 @@ class ClusterPage extends Component {
...
+                        SettingName.DEVELOPMENT.REDIRECT_TO_BETA_SWITCHED,
+                        NAMESPACES.DEVELOPMENT,
+                        true,
+                    );
+                    setSetting(
+                        SettingName.DEVELOPMENT.REDIRECT_TO_BETA,
+                        NAMESPACES.DEVELOPMENT,
+                        true,
+                    );
+                }
```
ma-efremoff, line 168: Please remove the `if`-statement and its comments.

packages/ui/src/ui/store/selectors/global/index.ts
ma-efremoff: `updateTitle` do nothing with settings, it cannot affect the issue.
There are many places where `getCluster` selector is use URL is not the best source of thruth.

ma-efremoff: let's move `updateTitle`-call up

packages/ui/src/ui/containers/ClusterPage/ClusterPage.js
ma-efremoff: Please do not touch the if, I will remove it by self later

ma-efremoff (approved):

Issue #370, opened 2024-02-20
Support validate and explain for YQL queries


Krock21:
After [server side support](https://github.com/ytsaurus/ytsaurus/issues/223) we need to add a convenient way to validate and explain(optimize) YQL queries

Button "Validate" should add "execution_mode: validate" to settings passed to start_query
Button "Explain" should add "execution_mode: optimize" to settings passed to start_query

zlobober:
Let's move ACO selector to the top of the page.

New buttons should appear next to "Run" button.

PR #430, opened 2024-03-20
feat(QueryTracker): support validate and explain for yql queries [#370]

vitshev:


packages/ui/src/ui/pages/query-tracker/QueryEditor/QueryEditor.tsx
```diff
@@ -156,35 +156,66 @@ const QueryEditorView = React.memo(function QueryEditorView({
...
-                        pin={isACOSupported ? 'round-brick' : undefined}
-                        className="query-run-action-button"
                         view="action"
                         onClick={runQueryCallback}
+                        className={b('action-button')}
                     >
                         <Icon data={playIcon} />
                         Run
                     </Button>
+                    {engine === QueryEngine.YQL ? (
```
SimbiozizV, line 201: I would simply combine 2 checks into one via `<><Button> ... </>`. But it's not critical

SimbiozizV (approved):

Issue #301, opened 2024-01-26
Add search field to statistics tab in Query Tracker


AMRivkin:
On the MapReduce operations page, under the statistics tab, there is a search field. Moreover, there are various aggregates. Additionally, there are hints and dimensions. I would like to add all of this to the Query Tracker.

<img width="1354" alt="image" src="https://github.com/ytsaurus/ytsaurus-ui/assets/7458598/0bfa9b8b-411d-4222-9f11-e6c5773daac2">

rBunin:
Make from that
![image](https://github.com/ytsaurus/ytsaurus-ui/assets/11693270/011cc3fe-e64d-4462-932d-6c9b40875eaa)

Make that

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/11693270/6c457ebe-7d3c-4e0e-8729-0210824493dd)

PR #438, opened 2024-03-25
feat(queryTracker): add search field to statistics tab in query tracker [#301]

vitshev:
https://nda.ya.ru/t/Ur3u3yt_76XxUM

### DIFF
--- a/packages/ui/src/ui/pages/job/tabs/Statistics/Statistics.tsx
+++ b/packages/ui/src/ui/pages/job/tabs/Statistics/Statistics.tsx
@@ -2,12 +2,14 @@ import * as React from 'react';
 import {useSelector} from 'react-redux';
 import {getRawStatistic} from '../../../../store/selectors/job/statistics';
 import {StatisticTable, StatisticTree} from '../../../../components/StatisticTable';
+import {isDocsAllowed} from '../../../../config';
+import UIFactory from '../../../../UIFactory';
 
 export default function Statistics() {
     const statistic = useSelector(getRawStatistic);
-
     return (
         <StatisticTable
+            helpUrl={isDocsAllowed() ? UIFactory.docsUrls['problems:jobstatistics'] : undefined}
             statistic={statistic as unknown as StatisticTree}
             visibleColumns={['min', 'max', 'last', 'sum', 'count']}
         />
--- a/packages/ui/src/ui/components/StatisticTable/StatisticTable.tsx
+++ b/packages/ui/src/ui/components/StatisticTable/StatisticTable.tsx
@@ -200,11 +200,13 @@ const useJobStatisticTable = ({
 };
 
 export function StatisticTable({
+    helpUrl,
     virtual,
     visibleColumns,
     fixedHeader,
     statistic,
 }: {
+    helpUrl?: string;
     virtual?: boolean;
     fixedHeader?: boolean;
     statistic: StatisticTree;
@@ -250,6 +252,7 @@ export function StatisticTable({
         <ErrorBoundary>
             <div className={block()}>
                 <Toolbar
+                    helpUrl={helpUrl}
                     onFilterChange={onFilterChange}
                     onTreeStateChange={setTreeState}
                     treeState={treeState}
--- a/packages/ui/src/ui/components/StatisticTable/Toolbar.tsx
+++ b/packages/ui/src/ui/components/StatisticTable/Toolbar.tsx
@@ -5,8 +5,6 @@ import cn from 'bem-cn-lite';
 import Filter from '../Filter/Filter';
 import HelpLink from '../HelpLink/HelpLink';
 
-import {isDocsAllowed} from '../../config';
-import UIFactory from '../../UIFactory';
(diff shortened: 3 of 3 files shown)


packages/ui/src/ui/pages/job/tabs/Statistics/Statistics.tsx
```diff
@@ -161,63 +61,44 @@ const tableProps = {
...
-            sum(item: JobTreeItem) {
-                return <Statistic item={item} aggregation="sum" />;
-            },
-            count(item: JobTreeItem) {
-                return <Statistic item={item} aggregation="count" />;
-            },
-        }),
-        [renderMetric],
-    );
+    const templates = {
```
ma-efremoff, line 67: It is better to move the templates to JobStatisticsTable component

packages/ui/src/ui/components/JobStatisticTable/JobStatisticTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+                        items={items}
+                        css={block()}
+                    />
+                </div>
+            </div>
+        </ErrorBoundary>
+    );
+}
+
+GeneralJobStatisticTable.Cell = JobStatisticTableStaticCell;
```
ma-efremoff, line 179: When it is possible it is better to avoid such "static" `exports` at least, separate exports are good enough and more clear. 
I think we shouldn't export `JobStatisticTableStaticCell` and `ExpandedCell`, instead of it we should describe `templates` in the file

packages/ui/src/ui/components/JobStatisticTable/JobStatisticTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+export type JobStatisticTableTemplate<Item extends Partial<JobTreeItem>> = {
+    [name: string]: (
+        item: Item,
+        colName: string,
+        toggleItemState: Function,
+        itemState: ItemState,
+    ) => React.ReactChild | null;
+};
+
+export function GeneralJobStatisticTable<Item extends Partial<JobTreeItem>>({
```
ma-efremoff, line 147: We don't need `Job` world in all named entities in the file, now it is just StatisticsTable

packages/ui/src/ui/components/JobStatisticTable/JobStatisticTable.tsx
```diff
@@ -0,0 +1,180 @@
...
+                    className={block('group-icon')}
+                />
+                <span>{renderValue(item)}</span>
+            </span>
+        );
+    }
+}
+
+export type TableProps<Item extends Partial<JobTreeItem>> = {
+    columns?: {
```
ma-efremoff, line 116: templates should be described in the component, and all we need instead of `TableProps` is a list of visible columns in properties of `JobStatisitcsTable`, something like:
```ts
visibleColumns: Array<'avg' | 'min' | 'max' | 'sum' | 'count' | 'last'>
```
We don't wan't any `ElementsTable`-specific properties for `GeneralJobStatisticTable`

packages/ui/src/ui/components/JobStatisticTable/JobStatisticTable.tsx
```diff
@@ -0,0 +1,180 @@
+import * as React from 'react';
+import {useCallback} from 'react';
+import cn from 'bem-cn-lite';
+
+import ErrorBoundary from '../ErrorBoundary/ErrorBoundary';
+import ElementsTableRaw from '../ElementsTable/ElementsTable';
+import Icon from '../Icon/Icon';
+import Toolbar from './Toolbar';
+
+import {JobTreeItem, LeafStatistic} from '../../types/job';
```
ma-efremoff, line 10: It would be better to avoid usage of Job-specific types. It looks like yql_statistics has the same format in response like job-statistics, so it would be better to handle raw-format in the component. We can React.useReducer and extract all necessary logic from corresponding reducers/actions/selectors.

packages/ui/src/ui/components/StatisticTable/prepare-job-statistic.ts
```diff
@@ -0,0 +1,29 @@
+import hammer from '../../common/hammer';
+import {JobStatistic, JobTree, MetricsEntry} from '../../types/job';
```
ma-efremoff, line 2: It should use types defined for `StatistcTable.tsx` instead of ones from `../../types/job`

packages/ui/src/ui/components/StatisticTable/prepare-job-statistic.ts
```diff
@@ -0,0 +1,29 @@
+import hammer from '../../common/hammer';
```
ma-efremoff, line 1: `prepare-job-statistic.ts` -> `prepare-statistic.ts`

ma-efremoff:
About query statistics:
1. Help link should no lead to jobs-statistics
2. There should be some visible border for sticky columns

ma-efremoff (approved):

PR #463, opened 2024-04-05
fix(Query): clique disabled by history query [YTFRONT-4105]

SimbiozizV:
https://nda.ya.ru/t/qkaNlvz975HzT8

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/module/query/actions.ts
+++ b/packages/ui/src/ui/pages/query-tracker/module/query/actions.ts
@@ -75,35 +75,6 @@ export const setCurrentClusterToQuery =
         dispatch(updateQueryDraft({settings: {...settings, cluster}}));
     };
 
-export function loadQuery(
-    queryId: string,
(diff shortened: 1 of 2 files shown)


packages/ui/src/ui/pages/query-tracker/module/query/actions.ts
```diff
@@ -138,6 +109,44 @@ export const loadCliqueByCluster =
...
+        try {
+            const query = await wrapApiPromiseByToaster(dispatch(getQuery(queryId)), {
+                toasterName: 'load_query',
+                skipSuccessToast: true,
+                errorTitle: 'Failed to load query',
+            });
+
+            if (
+                query.engine === QueryEngine.CHYT &&
+                query.settings &&
```
ma-efremoff, line 126: If we describe type of query.settings as `{cluster?: string} & Record<string, unknown>` we can use here:
```ts
if (query.engine === QueryEngine.CHYT && query.settings?.cluster) {
   ...
}
```

Issue #490, opened 2024-04-17
Error message displayed when entering incorrect login credentials


kaikash:
Hello!

When incorrect login credentials are provided the application displays the following error message: "Failed to show a login form. ytAuthCluster is not defined."

<img width="346" alt="image" src="https://github.com/ytsaurus/ytsaurus-ui/assets/8350069/4a69f8e7-89f6-41d2-a050-b5efb4548bab">

ma-efremoff:
@vitshev, another side effect of #349, please look at it.

PR #510, opened 2024-04-23
fix(login): error message displayed when entering incorrect login credentials [#490]

vitshev:


packages/ui/src/ui/entries/main.tsx
```diff
@@ -16,7 +16,11 @@ function onAxiosError(e: any) {
 }
 
 function onError(e: any) {
-    const isAuthError = axios.isAxiosError(e) && e.response?.status === 401;
+    const isAuthError =
+        axios.isAxiosError(e) &&
+        e.response?.status === 401 &&
+        e.response.data.message !== 'Invalid password';
```
ma-efremoff, line 22: It is bad idea to use `message` from response in such way. It is better to set some header from `server/controller/login` and check it here.

packages/ui/src/shared/constants/index.ts
```diff
@@ -5,6 +5,7 @@ export const ODIN_PAGE_ID = 'odin';
 
 export const YT_OAUTH_ACCESS_TOKEN_NAME = 'yt_oauth_access_token';
 export const YT_OAUTH_REFRESH_TOKEN_NAME = 'yt_oauth_refresh_token';
+export const YT_UI_CLUSTER_COOKIE_NAME = 'x-yt-ui-cluster-name';
```
ma-efremoff, line 8: YT_UI_AUTH_CLUSTER_HEADER_NAME

ma-efremoff (approved):

PR #514, opened 2024-04-23
refactor: migrate shards-related modules to typescript

vrozaev:


packages/ui/src/ui/store/thunkDispatch.ts
```diff
@@ -3,6 +3,6 @@ import {ThunkDispatch} from 'redux-thunk';
 import {RootState} from './reducers';
 import {useDispatch} from 'react-redux';
 
-export type AppThunkDispatch = ThunkDispatch<RootState, any, AnyAction>;
+export type AppThunkDispatch<T extends AnyAction = AnyAction> = ThunkDispatch<RootState, any, T>;
```
ma-efremoff, line 6: I think it is better to get rid of default value `= AnyAction` here, and set the parameter always explicitly.

packages/ui/src/ui/store/actions/components/shards.ts
```diff
@@ -72,11 +75,11 @@ export function openNameEditor(id) {
 export function closeNameEditor() {
     return {
         type: CLOSE_SHARD_NAME_EDITOR,
-    };
+    } as const;
```
ma-efremoff, line 78: Why do we need `as const` here?

ma-efremoff:
Please look up at the results of `Base / Lint & Unit ` check

packages/ui/src/ui/store/actions/components/shards.ts
vrozaev: Otherwise typescript loose exact type of CLOSE_SHARD_NAME_EDITOR and we got next error:

```
src/ui/store/actions/components/shards.ts:57:26 - error TS2769: No overload matches this call.
  Overload 1 of 3, '(thunkAction: ThunkAction<unknown, RootState, any, ShardsAction>): unknown', gave the following error.
    Argument of type '{ type: string; }' is not assignable to parameter of type 'ThunkAction<unknown, RootState, any, ShardsAction>'.
      Type '{ type: string; }' provides no match for the signature '(dispatch: ThunkDispatch<RootState, any, ShardsAction>, getState: () => RootState, extraArgument: any): unknown'.
  Overload 2 of 3, '(action: ShardsAction): ShardsAction', gave the following error.
    Argument of type '{ type: string; }' is not assignable to parameter of type 'ShardsAction'.
      Type '{ type: string; }' is not assignable to type 'Action<"COMPONENTS:SHARDS:CLOSE_SHARD_NAME_EDITOR">'.
        Types of property 'type' are incompatible.
          Type 'string' is not assignable to type '"COMPONENTS:SHARDS:CLOSE_SHARD_NAME_EDITOR"'.
  Overload 3 of 3, '(action: ShardsAction | ThunkAction<unknown, RootState, any, ShardsAction>): unknown', gave the following error.
    Argument of type '{ type: string; }' is not assignable to parameter of type 'ShardsAction | ThunkAction<unknown, RootState, any, ShardsAction>'.
      Type '{ type: string; }' is not assignable to type 'Action<"COMPONENTS:SHARDS:CLOSE_SHARD_NAME_EDITOR">'.

57                 dispatch(closeNameEditor());
```

ma-efremoff (approved):

PR #515, opened 2024-04-23
fix(Host): add ellipsis text in host

Flunt1k:
Fix rendering long host in dynamic tables

**Before**
Screenshot 2024-04-23 at 5 49 29 PM

**After**
Screenshot 2024-04-23 at 5 49 03 PM

### DIFF
--- a/packages/ui/src/ui/containers/Host/Host.scss
+++ b/packages/ui/src/ui/containers/Host/Host.scss
@@ -3,16 +3,10 @@
     overflow: hidden;
     align-items: center;
 
-    &__link {
-        display: inline-block;
+    &__tooltip {
         overflow: hidden;
         text-overflow: ellipsis;
         white-space: nowrap;
-        width: 100%;
-    }
-
-    &__tooltip {
-        width: 100%;
     }
 
     &_hidden {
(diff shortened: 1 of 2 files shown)


packages/ui/src/ui/containers/Host/Host.scss
```diff
@@ -3,6 +3,18 @@
     overflow: hidden;
     align-items: center;
 
+    &__link {
+        display: inline-block;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+        width: 100%;
```
ma-efremoff, line 11: It looks like we don't need any styles for `link` block, do we?

packages/ui/src/ui/containers/Host/Host.scss
```diff
@@ -3,6 +3,18 @@
...
+    }
+
+    &__tooltip {
+        width: 100%;
```
ma-efremoff, line 15: Shouldn't we set all required attributes with `text-overflow: ellipsis` here?
Do we really need width: 100%? Will it work correctly without it?

Flunt1k: You are right, we don't need to use styles on the link tag.

ma-efremoff (approved):

PR #520, opened 2024-04-25
feat(Navigation): sort button in table columns [YTFRONT-4135]

SimbiozizV:
https://nda.ya.ru/t/r1rrmQW075bx4y

### DIFF
--- a/packages/ui/src/ui/components/ColumnSelector/ColumnSelector.js
+++ b/packages/ui/src/ui/components/ColumnSelector/ColumnSelector.js
@@ -268,6 +268,19 @@ export default class ColumnSelector extends Component {
         });
     };
 
+    _handleDefaultSort = () => {
+        this.withActualItems(({items}) => {
+            return {
+                items: items.sort((a, b) => {
+                    const aCaption = a.caption || a.name;
+                    const bCaption = b.caption || b.name;
+
+                    return aCaption.localeCompare(bCaption);
+                }),
+            };
+        });
+    };
+
     _toggleShownItems = () => {
         this.setState((prevState) => ({
             showSelectedOnly: !prevState.showSelectedOnly,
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/components/ColumnSelector/ColumnSelector.js
```diff
@@ -268,6 +268,12 @@ export default class ColumnSelector extends Component {
         });
     };
 
+    _handleSort = () => {
+        this.withActualItems(({items}) => {
+            return {items: items.sort((a, b) => a.name.localeCompare(b.name))};
```
ma-efremoff, line 273: Default [`itemrRenderer`](https://github.com/ytsaurus/ytsaurus-ui/pull/520/files#diff-b85c197cdef3c085114f9f1a54aade6959549332b2c51af552d118e6be1aa0cfR134) uses `caption` field, so it is a bad idea to sort by `name`  field.
There is two ways to fix it:
1. We need additional property `sortAccessor` near `itemRenderer`, that will be used hear:
```ts
return {items: items.sort(a, b) => sortAccessor(a).localeCompare(sortAccessor(b))};
```
2. Get rid of `itemRenderer` property and mark `caption`  as required property of `items`.

SimbiozizV: corrected. "caption" is now taken into account

packages/ui/src/ui/components/ColumnSelector/ColumnSelector.js
```diff
@@ -268,6 +268,19 @@ export default class ColumnSelector extends Component {
         });
     };
 
+    _handleSort = () => {
```
evtaranov, line 271: nit: it would be nice to rename it to something like '_handleResetSort' or '_handleВуафгдеSort'

evtaranov (approved):
Looks good

evtaranov (approved):

PR #521, opened 2024-04-25
fix(Navigation): text bug in tablets layout [YTFRONT-4133]

SimbiozizV:
https://nda.ya.ru/t/9Lga7Bpf75byEH

### DIFF
--- a/packages/ui/src/ui/pages/navigation/tabs/Tablets/Tablets.js
+++ b/packages/ui/src/ui/pages/navigation/tabs/Tablets/Tablets.js
@@ -194,7 +194,14 @@ class Tablets extends Component {
 
     static renderHost(item) {
         const host = item.cell_leader_address;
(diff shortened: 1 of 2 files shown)


packages/ui/src/ui/components/ElementsTable/ElementsTable.scss
```diff
@@ -119,7 +119,6 @@ $thin-border: 1px solid var(--light-divider);
         padding: 0 8px;
 
         vertical-align: middle;
-        white-space: nowrap;
```
ma-efremoff, line 122: ElementsTable is used in many places of the project. Are you sure it won't break any other places?

SimbiozizV: double checked. I discovered that this change slightly affects the display of "Scheduling"

remade

ma-efremoff (approved):

PR #531, opened 2024-05-03
feat: added ability to render custom query tab at query-tracker page

vrozaev:


packages/ui/src/ui/UIFactory.tsx
```diff
@@ -355,6 +356,10 @@ export interface UIFactory {
 
     renderRolesLink(params: {cluster: string; login: string; className?: string}): React.ReactNode;
 
+    getCustomQueryTabName(): string;
+
+    renderCustomQueryTab(params: {query: QueryItem}): React.ReactNode;
```
ma-efremoff, line 361: Do we really need two separate methods? I think it is better to leave only one factory method, like:
```ts
getCustomQueryResultTab(): undefined | {title: string, renderContent() => React.ReactNode};
```

ma-efremoff (approved):

ma-efremoff (approved):

PR #532, opened 2024-05-07
feat(Query): vcs navigation [YTFRONT-4147]

SimbiozizV:
https://nda.ya.ru/t/DJN6ohO276FYAS
https://nda.ya.ru/t/_O9KRJrh75ncqW

### DIFF
--- a/packages/ui/src/server/components/vcs/index.ts
+++ b/packages/ui/src/server/components/vcs/index.ts
@@ -1,0 +1,61 @@
+import {VCSSettings} from '../../../shared/ui-settings';
+import {GithubApi} from './GithubApi';
+import {GitlabApi} from './GitlabApi';
+import {AxiosResponse} from 'axios';
+import {Request} from '@gravity-ui/expresskit';
+import {AppConfig} from '@gravity-ui/nodekit';
+import {ErrorWithCode} from '../../utils';
+
+export interface VcsApi {
+    token: string;
+    api: string;
+    vcsId: string;
+
+    getRequestHeaders(): {Authorization: string};
+    getRepositories(): Promise<unknown>;
+    getBranches(projectId: string): Promise<string[]>;
+    getRepositoryContent(projectId: string, branch: string, path: string): Promise<unknown[]>;
+    getFileContent(
+        projectId: string,
+        branch: string,
+        path: string,
+    ): Promise<AxiosResponse<NodeJS.ReadableStream>>;
+}
+
+export class VcsFactory {
+    static createVcs(vcs: VCSSettings, token?: string) {
+        if (!token) throw new ErrorWithCode(400, 'Token is required');
+
+        const apiConfig = {token, api: vcs.api, vcsId: vcs.id};
+        switch (vcs.type) {
+            case 'github':
+                return new GithubApi(apiConfig);
+            case 'gitlab':
+                return new GitlabApi(apiConfig);
+            default:
+                throw new ErrorWithCode(415, `Unsupported VCS type: ${vcs.type}`);
+        }
+    }
+
+    static getVcsSettings(req: Request): VCSSettings[] {
+        const {uiSettings} = req.ctx.config as AppConfig;
+
+        if (!('vcsSettings' in uiSettings))
+            throw new ErrorWithCode(404, 'vcsSettings is not found in UISettings');
+
+        return uiSettings.vcsSettings;
+    }
+
+    static getVcsSettingById(req: Request, vcsId?: string): VCSSettings {
+        if (!vcsId) throw new ErrorWithCode(400, 'vcsId is required');
+
+        const result = this.getVcsSettings(req).find((vcs) => vcs.id === vcsId);
+        if (!result)
+            throw new ErrorWithCode(
+                500,
+                'The current vcsId is missing from the vcsSettings setting in UISettings',
+            );
+
+        return result;
+    }
+}
--- a/packages/ui/src/server/controllers/vcs.ts
+++ b/packages/ui/src/server/controllers/vcs.ts
@@ -1,0 +1,198 @@
+import {AppRouteHandler, Response} from '@gravity-ui/expresskit';
+import axios from 'axios';
+import {VcsConfig, VcsRepository} from '../../shared/vcs';
+import {ErrorWithCode, sendError} from '../utils';
+import {VcsFactory} from '../components/vcs';
+
+const COOKIE_OPTIONS = {
+    httpOnly: true,
+    secure: true,
+    sameSite: 'strict' as const,
+    maxAge: 1000 * 60 * 60 * 24 * 365,
+};
+
+const ERROR_MESSAGE = {
+    VCS_CONFIG_NOT_FOUND: 'vcsSettings is not found in UISettings',
+    VCS_ID_REQUIRED: 'vcsId is required',
+    VCS_NOT_SUPPORTED: 'The current vcsId is missing from the vcsSettings setting in UISettings',
+    TOKEN_REQUIRED: 'Token is required',
+    PATH_REQUIRED: 'Path is required',
+    BRANCH_REQUIRED: 'Branch is required',
+    REPOSITORY_REQUIRED: 'Repository is required',
+    UNAUTHORIZED:
+        'The service cannot authorize you. Check the token. You can add new token in the section Settings -> VCS',
+};
+
+const cookieNameByVcsId = (vcsId?: string) => {
+    if (!vcsId) throw new ErrorWithCode(400, ERROR_MESSAGE.VCS_ID_REQUIRED);
+    return `vcs_${vcsId}`;
+};
+
+const sendApiError = (res: Response, error: unknown) => {
+    let status;
+    let message;
+
+    if (axios.isAxiosError(error)) {
+        status = error.response?.status;
+        message = error.response?.statusText;
+    } else if (error instanceof ErrorWithCode) {
+        status = error.code;
+        message = error.message;
+    } else if (error instanceof Error) {
+        message = error.message;
+    }
+
+    if (status === 401) {
+        status = 500;
+        message = ERROR_MESSAGE.UNAUTHORIZED;
+    }
+
+    sendError(res, message, status || 500);
+};
+
+export const getVcsConfig: AppRouteHandler = (req, res) => {
+    const cookies = req.cookies;
+    try {
+        const config = VcsFactory.getVcsSettings(req);
+        if (!config.length) throw new ErrorWithCode(404, ERROR_MESSAGE.VCS_CONFIG_NOT_FOUND);
+
+        const response = config.map<VcsConfig>((item) => ({
+            ...item,
+            hasToken: cookieNameByVcsId(item.id) in cookies,
+        }));
+        res.status(200).json(response);
+    } catch (e) {
+        req.ctx.logError('Error getting vcs configuration', e);
+        sendApiError(res, e);
+    }
+};
+
+export const createToken: AppRouteHandler = (req, res) => {
+    const {vcsId, token} = req.body;
+
+    try {
+        if (!token) throw new ErrorWithCode(400, ERROR_MESSAGE.TOKEN_REQUIRED);
+
+        const vcs = VcsFactory.getVcsSettingById(req, vcsId);
+        res.cookie(cookieNameByVcsId(vcs.id), token, COOKIE_OPTIONS);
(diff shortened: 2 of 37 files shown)


packages/ui/src/ui/pages/query-tracker/module/repoNavigation/selectors.ts
```diff
@@ -0,0 +1,33 @@
...
+    [selectRepository, selectRepositories, selectBranch, selectVcsConfig],
+    (repository, repositories, branch, config) => {
+        if (!repository) return '';
+
+        const currentRepository = repositories[repository];
+        const vcs = config.find(({id}) => id === currentRepository.vcsId);
+
+        if (!vcs) return '';
+
+        if (vcs.type === VcsType.GITHUB)
```
ma-efremoff, line 26: Is it possible to avoid such checks?

SimbiozizV: No. The paths are too different

ma-efremoff: Is it possible to declare some type like VscApi to incapsulate all VcsType-specific login in corresponding implementation?

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,320 @@
...
+    httpOnly: true,
+    secure: true,
+    sameSite: 'strict' as const,
+    maxAge: 1000 * 60 * 60 * 24 * 365,
+};
+
+const ERROR_MESSAGE = {
+    VCS_CONFIG_NOT_FOUND: 'Vcs config is not found',
+    VCS_ID_REQUIRED: 'Vcs id is required',
+    VCS_NOT_SUPPORTED: 'This vcs id is not supported',
```
ma-efremoff, line 19: As I understand `VCS_NOT_SUPPORTED` is used when a specific `vcsId` is not configured, so the message should contain something about configuration.

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    const config = loadVcsConfig(req);
+
+    if (!config.length) {
+        sendError(res, ERROR_MESSAGE.VCS_CONFIG_NOT_FOUND, 404);
+        return;
+    }
+
+    const vcs = config.find((item) => item.id === vcsId);
```
ma-efremoff, line 82:
```ts
const vcs = config?.find((item) => item.id === vcsId);
```
and you don't need to check `config.length`, and then `ERROR_MESSAGE.VCS_CONFIG_NOT_FOUND` is also unnecessary

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,320 @@
...
+    if (!vcs) {
+        sendError(res, ERROR_MESSAGE.VCS_NOT_SUPPORTED, 400);
+        return;
+    }
+
+    if (!token) {
+        sendError(res, ERROR_MESSAGE.TOKEN_REQUIRED, 400);
+        return;
+    }
```
ma-efremoff, line 91: It looks like all `vcsId`-specific checks might be placed in `VcsFactory.createVcs` instead of copy-pasting them for each method.

packages/ui/src/ui/pages/query-tracker/VcsNavigation/SettingsMenu/AddTokenForm.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const dispatch = useDispatch();
+    const [isSaving, setSaving] = useState(false);
+    const [vcs, setVcs] = useState<string | undefined>();
+    const [token, setToken] = useState('');
+    const config = useSelector(selectVcsConfig);
+
+    const handleSave = async () => {
+        if (vcs && token) {
+            setSaving(true);
+            await dispatch(saveToken(vcs, token));
```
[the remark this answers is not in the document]
ma-efremoff:
```ts
try {
    setSaving(true);
    await dispatch(saveToken(vcs, token));
} finally {
    setSaving(false);
    setVcs(undefined);
    setToken('');
}
```

packages/ui/src/ui/pages/query-tracker/QueriesList/index.tsx
```diff
@@ -47,18 +48,21 @@ export function QueriesList() {
     const dispatch = useDispatch();
     useEffect(() => {
         dispatch(requestQueriesList());
-    }, []);
+    }, [dispatch]);
+
+    const isNavigation = tab === QueriesListMode.VCS;
```
ma-efremoff, line 53: `isVsc`

packages/ui/src/ui/pages/query-tracker/module/vcsNavigation/actions.ts
```diff
@@ -0,0 +1,259 @@
+import {ThunkAction} from 'redux-thunk';
```
ma-efremoff, line 1: do we really need `Navigation` word in `.../module/vcsNavigation`?

packages/ui/src/ui/pages/query-tracker/module/vcsNavigation/actions.ts
```diff
@@ -0,0 +1,259 @@
...
+    setVcs,
+    setVcsConfig,
+} from './vcsNavigationSlice';
+import {selectVcsConfig, selectVcsNavigation} from './selectors';
+import {getQueryFiles} from '../query/selectors';
+import {updateQueryDraft} from '../query/actions';
+import guid from '../../../../common/hammer/guid';
+import {selectFileEditor} from '../queryFilesForm/selectors';
+import {setFileEditor} from '../queryFilesForm/queryFilesFormSlice';
+import {GithubRepository, GitlabRepository, VcsConfig} from '../../../../../shared/vcs';
```
ma-efremoff, line 26: Probably it is better to declare some type in `shared/vcs` with `GithubRepository | GitlabRepository`

packages/ui/src/server/controllers/vcs.ts
SimbiozizV: I don't really want to do another method in VcsFactory that doesn't relate to instantiation

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
+    try {
+        const config = VcsFactory.getVcsSettings(req);
+        if (!config.length) {
+            sendError(res, ERROR_MESSAGE.VCS_CONFIG_NOT_FOUND, 404);
+            return;
+        }
+
+        const response = config.map<VcsConfig>((item) => ({
+            ...item,
+            token: cookieNameByVcsId(item.id) in cookies,
```
ma-efremoff, line 50: It is better to avoid sending of tokens as is. But we can send only first and last letters like:
`some_long_user_token` -> `som****ken`

packages/ui/src/server/components/vcs/index.ts
```diff
@@ -0,0 +1,59 @@
...
+                return new GitlabApi(apiConfig);
+            default:
+                throw new Error(`Unsupported VCS type: ${vcs.type}`);
+        }
+    }
+
+    static getVcsSettings(req: Request): VCSSettings[] {
+        const {uiSettings} = req.ctx.config as AppConfig;
+        if (!('vcsSettings' in uiSettings)) {
+            throw new Error('vcsSettings is not found in UISettings');
```
ma-efremoff, line 44: It is better to use app-specific error type to send correct `http status code 400`.
```ts
// in server/utils/index.ts
export class BadRequest extends Error {}

export function sendError(res: Response, error: unknown, errCode = 500) {
    const errData = prepareErrorToSend(error);
    const statusCode = error instanceof BadRequest ? 400 : errCode;
    return res.set('content-type', 'application/json').status(statusCode).send(errData);
}

// here
throw new BadRequest(...)
```

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
+    } catch (e) {
+        req.ctx.logError('Error getting vcs configuration', e);
+        sendError(res, e, 500);
+    }
+};
+
+export const createToken: AppRouteHandler = (req, res) => {
+    const {vcsId, token} = req.body;
+
+    if (!vcsId) {
```
ma-efremoff, line 62: Corresponding exception will be thrown from `VcsFactory.getVcsSettingById`

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
+    return 'projectId' in repository
+        ? repository.projectId
+        : `${repository.owner}/${repository.name}`;
+};
+
+export const getRepositories: AppRouteHandler = async (req, res) => {
+    const cookies = req.cookies;
+    const vcsId = req.query.vcsId as string | undefined;
+
+    if (!vcsId) {
```
ma-efremoff, line 103: Corresponding exception will be thrown from VcsFactory.getVcsSettingById

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
+        sendError(res, ERROR_MESSAGE.VCS_ID_REQUIRED, 400);
+        return;
+    }
+
+    const token = cookies[cookieNameByVcsId(vcsId)];
+    if (!token) {
```
ma-efremoff, line 109: The check might be done in `VcsFactory.createVcs`

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
+    const cookies = req.cookies;
+    const repository = req.query.repository as VcsRepository | undefined;
+
+    if (!repository) {
+        sendError(res, ERROR_MESSAGE.REPOSITORY_REQUIRED, 400);
+        return;
+    }
+
+    const token = cookies[cookieNameByVcsId(repository.vcsId)];
+    if (!token) {
```
ma-efremoff, line 139: The check might be done in `VcsFactory.createVcs`

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
+        return;
+    }
+
+    if (!branch) {
+        sendError(res, ERROR_MESSAGE.BRANCH_REQUIRED, 400);
+        return;
+    }
+
+    const token = cookies[cookieNameByVcsId(repository.vcsId)];
+    if (!token) {
```
ma-efremoff, line 182: The check might be done in `VcsFactory.createVcs`

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
```
ma-efremoff, line 231: The check might be done in `VcsFactory.createVcs`

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
+        req.ctx.logError('Error create token', e);
+        sendError(res, e, 500);
+    }
+};
+
+export const removeToken: AppRouteHandler = (req, res) => {
+    const {vcsId} = req.body;
+
+    if (!vcsId) {
+        sendError(res, ERROR_MESSAGE.VCS_ID_REQUIRED, 400);
```
ma-efremoff, line 86: Missing `return` statement. Approach with throwing an exception excludes such misprints.

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -0,0 +1,261 @@
...
+    if (!token) {
+        sendError(res, ERROR_MESSAGE.TOKEN_REQUIRED, 400);
+        return;
+    }
+
+    try {
+        const vcs = VcsFactory.getVcsSettingById(req, vcsId);
+        const api = VcsFactory.createVcs(vcs, token);
+        const response = await api.getRepositories();
+        if (response instanceof AxiosError) {
```
ma-efremoff, line 118: Here and below methods of `api` never return instances of AxiosError. Such check should be mentioned once in sendApiError.

packages/ui/src/server/controllers/vcs.ts
SimbiozizV: const cookieNameByVcsId = (vcsId: string) => `vcs_${vcsId}`;

the token itself is not passed there. It looks like I should replace the variable name as it might be misleading

evtaranov (approved):
It looks like all critical issues have been resolved. Let's merge it.

PR #561, opened 2024-05-20
fix(Bundles): get list limits [YTFRONT-4177]

SimbiozizV:


packages/ui/src/ui/store/actions/chaos_cell_bundles/index.ts
```diff
@@ -43,13 +44,15 @@ export function fetchChaosBundles(): ChaosBundlesThunkAction {
                 parameters: {
                     path: '//sys/chaos_cells',
                     attributes: ['health', 'id', 'peers', 'status', 'tablet_cell_bundle'],
+                    max_size: USE_MAX_SIZE,
```
ma-efremoff, line 47: ...USE_MAX_SIZE
it looks like we need to describe max_size field for `list` in `BatchSubRequest`. And it is better to add explicit type for `requests`:
```ts
const requests: Array<BatchSubRequests> = [
    ...
]
```

packages/ui/src/ui/store/actions/chaos_cell_bundles/index.ts
```diff
@@ -43,13 +44,15 @@ export function fetchChaosBundles(): ChaosBundlesThunkAction {
...
                 },
             },
             {
                 command: 'get' as const,
                 parameters: {
                     path: '//sys/chaos_cell_bundles',
                     attributes: ['health', 'id', 'options'],
+                    max_size: USE_MAX_SIZE,
```
ma-efremoff, line 55: ...USE_MAX_SIZE

ma-efremoff (dismissed):

ma-efremoff (approved):

PR #571, opened 2024-05-21
feat(Components/Nodes): better sorting of progress columns [YTFRONT-3801]

ma-efremoff:
https://nda.ya.ru/t/xUb-rkJ5762U47


packages/ui/src/ui/components/ColumnHeader/ColumnHeader.tsx
```diff
@@ -33,58 +28,134 @@ export interface ColumnHeaderProps<T extends string = string> extends PageCounte
...
+    if (props.options === undefined) {
+        const {column, allowUnordered, withUndefined, title, shortTitle, allowedOrderTypes} = props;
+        return {
+            column,
+            allowUnordered,
+            withUndefined,
+            allowedOrderTypes,
+            ...getNameTitle({column, title, shortTitle}),
+        };
+    } else {
```
SimbiozizV, line 86: "else" is not needed. Above you do "return"

packages/ui/src/ui/components/ColumnHeader/ColumnHeader.tsx
```diff
@@ -100,3 +171,37 @@ export default function ColumnHeader<T extends string = string>(props: ColumnHea
...
+}) {
+    return !options ? null : (
+        <DropdownMenu
+            items={options.map((x) => {
+                const {titleContent} = getNameTitle(x);
+                return {
+                    text: titleContent,
+                    action() {
+                        const {column, allowUnordered, withUndefined, allowedOrderTypes} = x;
+                        const nextOrder = allowedOrderTypes
```
SimbiozizV, line 191: is it worth doing all these calculations if "onSort" may turn out to be "undefined"?

packages/ui/src/ui/pages/components/tabs/nodes/Nodes/NodesColumnHeader.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const dispatch = useDispatch();
+    const sortState = useSelector(getTables)[COMPONENTS_NODES_TABLE_ID];
+    const order = oldSortStateToOrderType(sortState);
+
+    const column = props.options?.find(({column}) => sortState.field === column);
+
+    return (
+        <ColumnHeader
+            {...props}
+            column={props.options ? column?.column ?? props.column : props.column}
```
SimbiozizV, line 26: double ternary operator. they seem so convenient but this is the path to evil

packages/ui/src/ui/store/selectors/tables.ts
```diff
@@ -0,0 +1,4 @@
+import {RootState} from '../../store/reducers';
+import {initialState} from '../../store/reducers/tables';
+
+export const getTables = (state: RootState) => state?.tables || initialState;
```
SimbiozizV, line 4: It's a little confusing. Why not process this in the store? We need to look at in what case "state?.tables" may not exist and why this is not processed in the action

packages/ui/src/ui/pages/components/tabs/nodes/Nodes/NodesColumnHeader.tsx
ma-efremoff: not double ternary operator

packages/ui/src/ui/store/selectors/tables.ts
ma-efremoff: the code was not changed it is just rewritted with typescript

packages/ui/src/ui/components/ColumnHeader/ColumnHeader.tsx
ma-efremoff: `SubColumnSelector` is displayed when `ColumnHeader.props.options` is defined and there is no reason to have `options` without `onSort`.

packages/ui/src/ui/components/ColumnHeader/ColumnHeader.tsx
ma-efremoff: Such isolated `if`-s (without `else`) looks like prevention of some unwanted behaviour below. But here `if/else`-expression is spliting of code-execution for two different types of input props of `HasSortColumn` (it is more like typeguard):
```ts
if (props.options === undefined) {
   // handling of type (ColumnInfo<T> & {options?: undefined})
} else {
  // handling of type 
  // ({
  //        title: string;
  //        shortTitle?: string;
  //        column: T;
  //        options: Array<ColumnInfo<T>>;
  //    } & Partial<Record<keyof Omit<ColumnInfo<T>, 'column' | 'title' | 'shortTitle'>, undefined>>)
}
```
The current form works the same as your suggestion but it is more expressive.

ma-efremoff: fixed

packages/ui/src/ui/store/selectors/tables.ts
ma-efremoff: fixed

PR #575, opened 2024-05-23
fix(Query): query and progress tab [YTFRONT-4185]

SimbiozizV:
https://nda.ya.ru/t/Ktcz1dbX7645Eb
 https://nda.ya.ru/t/369LYQ7Z76458u

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/module/query/actions.ts
+++ b/packages/ui/src/ui/pages/query-tracker/module/query/actions.ts
@@ -112,8 +112,10 @@ export const loadCliqueByCluster =
 
 export function loadQuery(
     queryId: string,
+    config?: {dontReplaceQueryText?: boolean},
 ): ThunkAction<any, RootState, any, SetQueryAction | RequestQueryAction | SetQueryErrorLoadAction> {
-    return async (dispatch, _getState) => {
+    return async (dispatch, getState) => {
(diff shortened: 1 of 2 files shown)


packages/ui/src/ui/pages/query-tracker/module/query/actions.ts
```diff
@@ -112,8 +112,10 @@ export const loadCliqueByCluster =
 
 export function loadQuery(
     queryId: string,
+    saveQuery?: boolean,
```
ma-efremoff, line 115: it is easier to read/refactor if we have name of such options in place of call:
```ts
loadQuery(id, {saveQuery: true});
```

packages/ui/src/ui/pages/query-tracker/module/query/actions.ts
```diff
@@ -128,10 +130,16 @@ export function loadQuery(
                 dispatch(loadCliqueByCluster(query.settings.cluster as string));
             }
 
+            const queryItem = prepareQueryPlanIds(query);
+
+            if (saveQuery) {
```
ma-efremoff, line 135: The name of the parameter is confusing. It looks like `loadQuery` ignores response when `saveQuery === true`, doesn't it?

ma-efremoff (approved):

Issue #533, opened 2024-05-07
Utf-8 in QT


spiner2000:
If you got field name in utf-8 you will have "\<unsupported data type\>" in query output values
![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1374700/a00a9081-a17c-4f5a-afcc-2b101648c16b)
But if alias is set 
![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1374700/b7af3d9b-1d9e-4d0e-803a-5043711d8b74)
it works fine
![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1374700/38bae7e1-93b0-41d0-a7d1-0578b4fec08f)

SimbiozizV:
The problem will be resolved as soon as https://github.com/ytsaurus/ytsaurus-ui/pull/597 will be added to the prod

SimbiozizV:
fix in main

PR #597, opened 2024-05-30
fix(Query): utf decode in result table [#533]

SimbiozizV:
https://nda.ya.ru/t/8PLJRXyu76AqvQ

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/module/api.ts
+++ b/packages/ui/src/ui/pages/query-tracker/module/api.ts
@@ -160,6 +160,14 @@ export const CompletedStates = [
     QueryStatus.FAILED,
 ];
 
+const secureDecoding = (value: string) => {
+    try {
+        return unipika.decode(value);
+    } catch (e) {
+        return value;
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/pages/query-tracker/module/api.ts
```diff
@@ -167,13 +167,15 @@ const JSONParser = {
...
-                if (typeof value === 'string') {
-                    try {
-                        return unipika.decode(value);
-                    } catch (e) {
-                        return value;
-                    }
+                if (typeof value === 'object' && value !== null && !Array.isArray(value)) {
+                    return Object.keys(value).reduce<Record<any, any>>((acc, k) => {
+                        acc[unipika.decode(k)] =
+                            typeof value[k] === 'string' ? unipika.decode(value[k]) : value[k];
```
ma-efremoff, line 173: `unipika.decode` may throw an exception.

ma-efremoff (approved):

PR #607, opened 2024-06-05
fix(Navigation): widget with footer problem [YTFRONT-4221]

SimbiozizV:
https://a.yandex-team.ru/review/6038384/files/1


packages/ui/src/ui/containers/ClusterPage/ClusterPage.scss
```diff
@@ -10,6 +10,10 @@
         }
     }
 
+    &__panes-wrapper_with-footer {
+        max-height: calc(100vh - var(--app-header-height) - 40px);
```
ma-efremoff, line 14: Who can guarantee `40px` height of footer?

PR #616, opened 2024-06-14
chore(tests): add screenshots for Operations page

ma-efremoff:
https://nda.ya.ru/t/c-avrXs276RTt9


packages/ui/tests/screenshots/pages/navigation/NavigationPage.ts
```diff
@@ -44,6 +44,14 @@ export class NavigationPage extends BasePage {
     async waitForACL() {
         await this.page.waitForSelector('.navigation-acl__row .yt-subject-link');
     }
+
+    async replaceDyntableSize() {
+        await replaceInnerHtml(this.page, {
+            '.map-nodes-table__row_dyntable .map-node_default__table-item_type_disk-space':
+                '000000 KiB',
+            '.map-nodes-table__row_dyntable .map-node_default__table-item_type_rows': '00000',
```
SimbiozizV, line 52: Ideally we shouldn't have such manual substitutions. We have to mock the input data and take a screenshot from it. And do not replace what you want with artificial data.

Binding to classes is dangerous because we tightly bind tests to the structure. Any change in the structure or names of classes will require regeneration. Although there may be no visual changes.

Is it possible to get rid of these substitutions?

SimbiozizV (approved):

packages/ui/tests/screenshots/pages/navigation/NavigationPage.ts
ma-efremoff: Discussed offline.

PR #617, opened 2024-06-17
feat(Query): spyt clicue selector [YTFRONT-4219]

SimbiozizV:
https://nda.ya.ru/t/ymgnP5Dy76TgUL


packages/ui/src/server/controllers/strawberry-api.ts
```diff
@@ -32,29 +32,38 @@ async function chytProxyApiImpl(req: Request, res: Response) {
         'get_speclet',
     ]);
 
+    if (!engine) {
```
ma-efremoff, line 35: It is better to check such parameters for specific allowed values
```ts
const ENGINES = {
    chyt: 'chyt_controller_base_url',
    spyt: 'livy_controller_base_url',
}
```
Then you don't need `isChyt` variable.

packages/ui/src/ui/utils/strawberryControllerApi.ts
```diff
@@ -110,6 +111,7 @@ type StrawberryApiType = <
     cluster: string,
     params: ApiItem['params'],
     props: {
+        type?: QueryEngine.CHYT | QueryEngine.SPYT;
```
ma-efremoff, line 114: type?: 'chyt' | 'spyt'

SimbiozizV: I don’t understand why you don’t like enum. Now it is done more centrally

packages/ui/src/server/controllers/strawberry-api.ts
SimbiozizV: ENGINES is not a suitable name. I did it a little differently

packages/ui/src/ui/utils/strawberryControllerApi.ts
ma-efremoff: The enum contains unnecessary values that not supported by the api

ma-efremoff (approved):

packages/ui/src/ui/pages/query-tracker/QueryTrackerTopRow/index.tsx
```diff
@@ -39,15 +39,23 @@ const QueryTrackerTopRow: FC = () => {
...
-                dispatch(loadCliqueByCluster(settings.cluster as string));
+            const isSpyt = newEngine === QueryEngine.SPYT;
+            const isChyt = newEngine === QueryEngine.CHYT;
+
+            if ((isSpyt || isChyt) && settings && 'cluster' in settings) {
+                dispatch(
+                    loadCliqueByCluster(
+                        newEngine === QueryEngine.SPYT ? QueryEngine.SPYT : QueryEngine.CHYT,
+                        settings.cluster as string,
+                    ),
```
ma-efremoff, line 50: it might be a little simplified
```ts
if ((isSpyt || isChyt) && settings?.cluster) {
    dispatch(loadCliqueByCluster(newEngine, settings.cluster))
}
```

packages/ui/src/ui/pages/query-tracker/QueryTrackerTopRow/index.tsx
```diff
@@ -78,8 +86,8 @@ const QueryTrackerTopRow: FC = () => {
             const newSettings: Record<string, string> = settings ? {...settings} : {};
             if (clusterId) {
                 newSettings.cluster = clusterId;
-                if (engine === QueryEngine.CHYT) {
-                    dispatch(loadCliqueByCluster(clusterId));
+                if (engine === QueryEngine.CHYT || engine === QueryEngine.SPYT) {
+                    dispatch(loadCliqueByCluster(engine, clusterId));
```
ma-efremoff, line 90: At the momemnt the action is called from 3 different places:
- handleClusterChange
- handleChangeEngine
- loadCliqueByCluster

They all might be replaced by one React.useEffect:
```ts
React.useEffect(() => {
    if (...) {
        dispatch(loadCliqueByCluster(engine, clusterId));
    }
    ...
}, [engine, cluster])
```

SimbiozizV: Used in 2 locations: handleClusterChange и handleChangeEngine
I don't see the point in putting this in a separate useEffect. In the current implementation, we get new data when calling handle and immediately enter it in store

useEffect is triggered after the component has rendered. In other words, we will enter some of the data, the component will pull back and useEffect will work, which will add data via loadCliqueByCluster.

Does this edit optimize anything for us? I don't think

ma-efremoff: I have misprint in my message `loadCliqueByCluster` -> `loadQuery`

ma-efremoff (approved):

PR #621, opened 2024-06-18
feat: added ability to show custom tab in navigation view

vrozaev:
https://nda.ya.ru/t/atP5j4Lb76WCss
With this PR, custom yt installations can add their own tabs for navigation.

Here is an example - we added a new tab based on attributes:

Screenshot 2024-06-18 at 10 31 29

### Factory overrides:

``` typescript
const productionOverrides: Partial<UIFactory> = {
    getNavigationExtention() {
        const TAB_KEY = 'cutsom_tab_with_custom_tab';
        return {
            extendSupportedTabs({supportedTabs, attributes}) {
                if (attributes.ui_flag_for_custom_ui) {
                    supportedTabs.add(TAB_KEY);
                }
            },

            getDefaultMode(supportedTabs: Set<string>) {
                if (supportedTabs.has(TAB_KEY)) {
                    return TAB_KEY;
                }
                return supportedTabs.has(NavigationTab.CONSUMER)
                    ? NavigationTab.CONSUMER
                    : NavigationTab.CONTENT;
            },

            getSupportedAttributeTypes() {
                function Stub() {
                    return <div>Here will be custom react component</div>;
                }

                return {
                    [TAB_KEY]: Stub,
                };
            },
            extendNavigationTabs(tabs) {
                tabs.unshift({
                    value: TAB_KEY,
                    title: 'Go to custom tab',
                });
            },
        };
    },
};
```

### DIFF
--- a/packages/ui/src/ui/UIFactory.tsx
+++ b/packages/ui/src/ui/UIFactory.tsx
@@ -99,6 +99,19 @@ export type AclRoleActionsType = Partial<Omit<PreparedAclSubject | PreparedRole,
     type?: string;
 };
 
+export type TabName = string;
+
+export type ExtraTab = {
+    value: `extra_${TabName}`;
+    title: string;
+    text?: string;
+    caption?: string;
+    hotkey?: string;
+    component: React.ComponentType;
+    isSupported: (attributes: Record<string, any>) => boolean;
+    position: {before: TabName} | {after: TabName};
+};
+
 export interface UIFactory {
     getClusterAppearance(cluster?: string): undefined | ClusterAppearance;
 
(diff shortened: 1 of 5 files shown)


packages/ui/src/ui/UIFactory.tsx
```diff
@@ -389,6 +390,16 @@ export interface UIFactory {
     getAclPermissionsSettings(): typeof PERMISSIONS_SETTINGS;
 
     onChytAliasSqlClick(params: {alias: string; cluster: string}): void;
+
+    getNavigationExtention(): {
```
ma-efremoff, line 394: It looks like it is enough to describe additional tabs in declarative way, like:
```ts
type TabName = string;
type ExtraTab = {
    value: `extra_${TabName}`,
    title: string;
    component: React.ComponentType;
    position?: {before: TabName} | {after: TabName}
}
...
getNavigationExtraTabs(): undefined | Array<ExtraTab>;
```

vrozaev: Done

ma-efremoff (approved):

Issue #636, opened 2024-06-21
Customizable welcome messages on the login page


zlobober:
Currently, login page looks like the following:

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/df619c48-2d53-42e3-a710-cb3f1a8a5c28)

On installations with basic password-based authentication it is typical to have some sort of an external password management, e.g. some 3rd party service creates users, issues secrets (password, token), and stores them in some lockbox/vault-like secret storage.

The common issue is how to notify users, where to look for their password. In order to make this simpler, I'd suggest having a customizable text on this page, i.e. putting the constant

```
A unified scalable platform for storing and processing large volumes of data.
Login to your account.
```

into a markdown-formatted variable in the UI config.

In my company I would replace this text with something like:

```
Log in using your credentials for cluster <clustername>,
which may be obtained by the instruction by the link <link>.
```

In order to make the link clickable, the text must be markdown-formatted.

PS Also, the default text is misleading, as the word "Account" means different thing in YT. I'd suggest using the wording "Login under your credentials."

PR #645, opened 2024-07-02
feat: added ability to override default text on login page

vrozaev:
https://github.com/ytsaurus/ytsaurus-ui/issues/636

Now we can change default text on login page via ui settings.

image


packages/ui/src/shared/ui-settings.ts
```diff
@@ -146,6 +146,14 @@ export interface UISettings {
      * @example reUseEffectiveAclForPath: '//sys/access_control_object_namespaces[^/+]{0,}'
      */
     reUseEffectiveAclForPath?: string;
+
+    /**
+     * Allows to override default title and text on login page, accepts HTML string.
+     */
+    loginPageSettings?: {
```
vrozaev, line 153: Should I put this logic to UIFactory 🤔?

PR #657, opened 2024-07-10
feat(Query): share query button [YTFRONT-4239]

SimbiozizV:
https://nda.ya.ru/t/CKdx2QDq76stEf

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/module/query_aco/selectors.ts
+++ b/packages/ui/src/ui/pages/query-tracker/module/query_aco/selectors.ts
@@ -2,7 +2,9 @@ import intersection from 'lodash/intersection';
 import {SelectOption} from '@gravity-ui/uikit/build/esm/components/Select/types';
 import {RootState} from '../../../../store/reducers';
 import {getSettingsData} from '../../../../store/selectors/settings-base';
-const getState = (state: RootState) => state.queryTracker.aco;
+import {createSelector} from 'reselect';
+import {SHARED_QUERY_ACO} from '../query/selectors';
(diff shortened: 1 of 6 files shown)


packages/ui/src/ui/pages/query-tracker/module/query_aco/selectors.ts
```diff
@@ -24,6 +26,19 @@ export const getQueryACOOptions = (state: RootState): SelectOption[] => {
...
+    getState(state).data.access_control_objects;
+
 export const isQueryTrackerInfoLoading = (state: RootState) => getState(state).loading;
 
 export const getQueryTrackerInfo = (state: RootState) => getState(state).data;
+
+export const isSupportedShareQuery = createSelector(
+    [selectIsMultipleAco, selectAccessControlObjects],
+    (isMultipleAco, aco) => {
+        return Boolean(isMultipleAco) && aco.includes(SHARED_QUERY_ACO);
```
ma-efremoff, line 42: `aco.includes(SHARED_QUERY_ACO)`-part looks suspiciously here.
Doesn't it mean such query is shared? If so it should not affect visibility of `Share`/`Unshare` button.

SimbiozizV: the fact is that we are moving from a regular aco field to an aco array. Both variants of this field are currently being sent in the request. As a string and as an array. At the moment, the array has a higher priority. So I check for a specific rule in the array

ma-efremoff (approved):

ma-efremoff (approved):

PR #660, opened 2024-07-12
feat(Query): path autocomplete [YTFRONT-4264]

SimbiozizV:
https://nda.ya.ru/t/Gz-ctXB_76vBxQ

### DIFF
--- a/packages/ui/src/ui/libs/monaco-yql-languages/helpers/getDirectoryContent.ts
+++ b/packages/ui/src/ui/libs/monaco-yql-languages/helpers/getDirectoryContent.ts
@@ -1,0 +1,71 @@
+import {YTApiId, ytApiV3, ytApiV3Id} from '../../../rum/rum-wrap-api';
+import {IRange, Position, editor, languages} from 'monaco-editor';
+import ypath from '../../../common/thor/ypath';
+import {YPathItem} from '../../../pages/query-tracker/Plan/services/tables';
+import {getPathFromMonacoModel} from './getPathFromMonacoModel';
+import {QueryEngine} from '../../../pages/query-tracker/module/engines';
+import {SuggestionsWeight} from './generateSuggestions';
+import {suggestionIndexToWeight} from './suggestionIndexToWeight';
+import {getClusterProxy} from '../../../store/selectors/global';
+import YT from '../../../config/yt-config';
+
+type Props = (data: {
+    model: editor.ITextModel;
+    monacoCursorPosition: Position;
+    engine: QueryEngine;
+    range: IRange;
+}) => Promise<languages.CompletionItem[]>;
+
+export const getDirectoryContent: Props = async ({model, monacoCursorPosition, engine, range}) => {
+    const {path, cluster} = getPathFromMonacoModel(model, monacoCursorPosition, engine);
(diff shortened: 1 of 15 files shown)


packages/ui/src/ui/libs/monaco-yql-languages/helpers/getDirectoryContent.ts
```diff
@@ -0,0 +1,72 @@
...
+        let clearPath = path.replace(/(\/$)/g, '');
+        let pathExist = await ytApiV3.exists({path: clearPath});
+        if (!pathExist) {
+            if (/[^/]\/$/.test(path)) return []; // we exclude the situation when we have the wrong path
+            clearPath = clearPath.split('/').slice(0, -1).join('/');
+            pathExist = await ytApiV3.exists({path: clearPath});
+            if (!pathExist) return [];
+        }
+
+        const clusterConfig = cluster && cluster in YT.clusters ? YT.clusters[cluster] : null;
```
ma-efremoff, line 33: `cluster` should be always defined as a correct value else the function should not be called.
If `cluster` is not mentioned then a query is started on QT-cluster instead of current one, isn't it?

packages/ui/src/ui/libs/monaco-yql-languages/helpers/getDirectoryContent.ts
```diff
@@ -0,0 +1,72 @@
...
+
+type Props = (data: {
+    model: editor.ITextModel;
+    monacoCursorPosition: Position;
+    engine: QueryEngine;
+    range: IRange;
+}) => Promise<languages.CompletionItem[]>;
+
+export const getDirectoryContent: Props = async ({model, monacoCursorPosition, engine, range}) => {
+    const {path, cluster} = getPathFromMonacoModel(model, monacoCursorPosition, engine);
```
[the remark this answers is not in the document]
SimbiozizV: Now the logic is as follows: 
1) if the request body specifies use or chuster_name before`, then we take the value from these lines
2) if the request body doesn't contain use or chuster_name. Then we take the main cluster and not from the selector
This is consistent logic. We don't take the value from the selector

packages/ui/src/ui/libs/monaco-yql-languages/helpers/getDirectoryContent.ts
SimbiozizV: if the cluster is not specified in the request body, then we use the structure of the current (main)cluster

packages/ui/src/ui/libs/monaco-yql-languages/helpers/getDirectoryContent.ts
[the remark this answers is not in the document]
SimbiozizV: I don't see anything wrong with the function knowing about Monaco. After all, this is the autocomplete event handler in the editor

packages/ui/src/ui/libs/monaco-yql-languages/helpers/getDirectoryContent.ts
SimbiozizV: returning an empty sagest if there is no cluster

ma-efremoff (approved):

PR #664, opened 2024-07-16
minor fixes

ma-efremoff:
https://nda.ya.ru/t/zhtoJh3e76zRqd

### DIFF
--- a/packages/ui/src/ui/components/PreloaderErrors/GeneralErrorRegion.js
+++ b/packages/ui/src/ui/components/PreloaderErrors/GeneralErrorRegion.js
@@ -12,7 +12,9 @@ import {showErrorPopup} from '../../utils/utils';
 
 const b = block('preloader');
 
-export function GeneralErrorRegion({message}) {
+export function GeneralErrorRegion({
+    message = 'Oops! Something went wrong. If problem persists please report it via Bug Reporter.',
+}) {
     const error = useSelector(getGlobalError);
 
     return (
@@ -39,13 +41,7 @@ export function GeneralErrorRegion({message}) {
 }
 
 GeneralErrorRegion.propTypes = {
-    cluster: PropTypes.string,
     message: PropTypes.string,
-};
-
-GeneralErrorRegion.defaultProps = {
-    cluster: 'Cluster',
-    message: 'Oops! Something went wrong. If problem persists please report it via Bug Reporter.',
 };
 
 export default withBlockedNavigation(GeneralErrorRegion);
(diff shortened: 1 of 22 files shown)


packages/ui/src/ui/components/PreloaderErrors/GeneralErrorRegion.js
```diff
@@ -12,7 +12,9 @@ import {showErrorPopup} from '../../utils/utils';
 
 const b = block('preloader');
 
-export function GeneralErrorRegion({message}) {
+export function GeneralErrorRegion({
+    message = 'Oops! Something went wrong. If problem persists please report it via Bug Reporter.',
```
SimbiozizV, line 16: I would put the above into a constant. DEFAULT_ERROR_TEXT = 'Oops... '. It would be a bit cleaner. The text is still long

packages/ui/src/ui/components/formatters/FormattedLink.js
```diff
@@ -7,7 +7,7 @@ import Link from '../../components/Link/Link';
 import FormattedText from './FormattedText';
 
 function FormattedLink(props) {
-    const {state, theme, className, text, match, onClick, ...rest} = props;
+    const {state, theme = 'ghost', className, text, match, onClick, ...rest} = props;
```
SimbiozizV, line 10: I didn't quite understand the approach. In packages/ui/src/ui/components/OperationPool/OperationPool.tsx, you transfer destructuring assignment to the attributes of the function, but not here

packages/ui/src/ui/store/actions/components/nodes/nodes.ts
```diff
@@ -86,17 +86,27 @@ export function getNodes({
...
+
+                for (const output of outputs) {
+                    for (const item of output) {
+                        const {$value: name} = item;
+                        if (!collectedItems.has(name)) {
+                            nodes.push(item);
+                            collectedItems.add(name);
+                        }
+                    }
+                }
```
SimbiozizV, line 106: now the complexity of the algorithm is O (n^2). Perhaps here you can make outputs.flat and delete duplicates via Set

packages/ui/src/ui/components/formatters/FormattedLink.js
ma-efremoff: In OperationPool.tsx the line prettified to several lines, but not here.

packages/ui/src/ui/components/PreloaderErrors/GeneralErrorRegion.js
ma-efremoff: Until the text is reused we don't need to extract the text to a separate value.

SimbiozizV (approved):

Issue #522, opened 2024-04-25
Redirect to YT operations from running YQL queries


Krock21:
Server-side issue: https://github.com/ytsaurus/ytsaurus/issues/551
Previous UI issue: https://github.com/ytsaurus/ytsaurus-ui/issues/369

Currently, redirect to YT operations in YQL query only works after query is finished
We need to be able to do it while a query is running

Krock21:
Server-side issue completed. This issue is ready to be taken

Krock21:
yql_progress now contains this information for every node:
```
"remoteData": {
    "cluster_name": "dirac"
},
```
You may use it to form a correct link to an operation

Please preserve current working state with completed operations and old server version (keep looking at `yql_statistics.yt.[id]. _cluster_name` and using it, if `yql_progress.[id].remoteData.cluster_name` is not available)

PR #681, opened 2024-07-22
fix(QueryTracker): Redirect to YT operations from running YQL queries [#522]

vitshev:


packages/ui/src/ui/pages/query-tracker/Plan/utils.ts
```diff
@@ -596,9 +596,10 @@ export function usePrepareNode(operationIdToCluster: Map<string, string>) {
         } else if (node.progress?.remoteId) {
             const id = node.progress?.remoteId.split('/').pop();
 
-            if (id && operationIdToCluster.has(id)) {
-                const cluster = operationIdToCluster.get(id)!;
-
+            const cluster = operationIdToCluster.has(id!)
```
SimbiozizV, line 599: The id can be undefined. Now the check is disabled via !
I think it's better to add a check `if(!id) {node.url = getOperationUrl(node); return node;}`

ma-efremoff (approved):

PR #684, opened 2024-07-23
fix(MaintenancePage): fix for maintenance page

ma-efremoff:


packages/ui/src/ui/hocs/withBlockedNavigation.js
```diff
@@ -33,3 +30,16 @@ export default function withBlockedNavigation(Component) {
 
     return connect(null, {addNavigationBlocker, removeNavigationBlocker})(ResComponent);
 }
+
+export function BlockNavigation() {
```
SimbiozizV, line 34: At its core, it is not a component. Because it doesn't have any visual component. I suggest moving this to an action or hook

packages/ui/src/ui/containers/MaintenancePage/HandleMaintenance.tsx
```diff
@@ -1,62 +1,51 @@
...
+    children: React.ReactNode;
+    maintenanceContent?: React.ReactNode;
+};
 
+type ReduxProps = ConnectedProps<typeof connector>;
+
+class HandleMaintenance extends React.Component<Props & ReduxProps> {
     render() {
-        const {location, cluster, maintenancePageEvent} = this.props;
+        const {cluster, maintenancePageEvent, maintenanceContent, children} = this.props;
```
SimbiozizV, line 19: I would do this `if (! maintenancePageEvent || isIgnored) return children; ` this would save the code from 1 level of nesting

packages/ui/src/ui/hocs/withBlockedNavigation.js
ma-efremoff: It is impossible to use hooks from class-components, so we don't have any choice here.

SimbiozizV (approved):

Issue #700, opened 2024-08-01
Add ability to disable filter optimisation on operations page


vrozaev:
![image](https://github.com/user-attachments/assets/554e6594-df7c-407c-8521-009e6b49ffbf)

PR #701, opened 2024-08-01
feature: now we can disable filter optimization on the operations page via cluster config [#700]

vrozaev:


packages/ui/src/ui/pages/operations/OperationsList/OperationsListToolbar/OperationsListToolbar.js
```diff
@@ -222,7 +222,7 @@ class OperationsListToolbar extends React.PureComponent {
 
     renderWarning_uiissue_2838() {
         return (
-            <div className={block('ytfront-2383')}>
+            <div className={block('ytfront-2838')}>
```
vrozaev, line 225: I changed css class here because the issue numbers are different. I guess that ytfront-2838 is the right one and ytfront-2383 is just a typo?

packages/ui/src/shared/yt-types.d.ts
```diff
@@ -108,6 +108,10 @@ export interface ClusterConfig {
         title?: string;
         text?: string;
     };
+
+    operationPageSettings?: {
+        disableOptimizationForYTFRONT2838: boolean;
```
vrozaev, line 113: I'm open for suggestions regarding the naming.

ma-efremoff (approved):

Issue #702, opened 2024-08-02
A usage "view" button for truncated cells in the QueryTracker table


Flunt1k:
Use case:

User got a table with cells which are `{inc: true, data: ""}`. Cells may contain json/yson/image or text.
User want to see cell data.

Related to the [issue](https://github.com/ytsaurus/ytsaurus-ui/issues/655), we want to add preview button to QT table as well.

PR #709, opened 2024-08-06
feat(YQLTable): add "view" button for truncated cells [#702]

Flunt1k:


packages/ui/src/ui/components/CellPreviewModal/CellPreviewModal.tsx
```diff
@@ -3,23 +3,23 @@ import React from 'react';
 import {useDispatch, useSelector} from 'react-redux';
```
ma-efremoff, line 3: React-components linked with store we put to `src/ui/containers`, and `src/ui/components` should contain components independed from store.

packages/ui/src/ui/components/CellPreviewModal/CellPreviewModal.tsx
```diff
@@ -45,26 +45,32 @@ export const CellPreviewModal: React.FC = () => {
...
-                    </Text>
-                    <code className={b('command-wrapper')}>
-                        <div className={b('command')}>
-                            <div className={b('command-content')} title={command}>
-                                {command}
+                {command ? (
+                    <Flex gap={2} direction="column">
+                        <Text variant="subheader-3" color="secondary">
+                            {data?.$incomplete && !error
+                                ? 'Unable to load content more than 16MiB. Please use the command bellow to load it locally.'
```
ma-efremoff, line 52: Is it correct message for QT results?

packages/ui/src/ui/pages/query-tracker/QueryResultsView/YQLTable/YQLTable.tsx
```diff
@@ -73,6 +76,12 @@ type Props = {
     onSort?: (sortOrder?: SortOrder | SortOrder[]) => void;
     showFullValueInDialog?: boolean;
     containerRef?: React.Ref<HTMLDivElement>;
+    queryParams: TableQueryParams;
```
ma-efremoff, line 79: Do we really need to know about `queryParams` on this level? Isn't it better to pass callback `onShowPreveview`?
```ts
onShowPreview: (colName: string, rowIndex: number) => void;
```

packages/ui/src/ui/pages/query-tracker/QueryResultsView/YQLTable/YQLTable.tsx
```diff
@@ -407,13 +453,20 @@ function unquote(value?: string) {
...
+const cellActionClassName = block('cell-action');
 
 interface TableCellProps {
     children?: StrictReactNode;
     rawValue?: string;
+    onPreviewClick: () => void;
+    isTruncated?: boolean;
+}
+
+function isRawValue(value: string | undefined): value is string {
```
ma-efremoff, line 465: What is the reason to have such type-guard function? 
Isn't it much clear to use `typeof value === string`/`Boolean(value)` check in place of usage?

packages/ui/src/ui/containers/CellPreviewModal/CellPreviewModal.tsx
```diff
@@ -47,24 +48,30 @@ export const CellPreviewModal: React.FC = () => {
             <Flex className={b('content')} gap={2} direction="column">
                 <Flex gap={2} direction="column">
                     <Text variant="subheader-3" color="secondary">
-                        {data?.$incomplete && !error
-                            ? 'Unable to load content more than 16MiB. Please use the command bellow to\n' +
-                              '                            load it locally.'
-                            : 'You could use the command bellow to load it locally.'}
+                        {incompleteNoticeText && incompleteNoticeText}
```
ma-efremoff, line 51: Looks like a misprint:
```ts
{incompleteNoticeText && incompleteNoticeText}
// it looks like it should be replaced with
{incompleteNoticeText}
// if so the field should be renamed to `noticeText`
// cause we don't need to check `data?.$incomplete` here
```

packages/ui/src/ui/store/actions/navigation/modals/cell-preview.ts
```diff
@@ -61,13 +31,19 @@ export const showCellPreviewModal = (columnName: string, index: number): ActionT
...
             });
 
             const parsed = JSON.parse(json);
 
             const data = parsed.rows[0][columnName];
-            dispatch({type: CELL_PREVIEW.SUCCESS, data: {data}});
+
+            const isIncomplete = data.$incomplete;
+            const incompleteNoticeText = isIncomplete
+                ? 'Unable to load content more than 16MiB. Please use the command bellow to load it locally.'
```
ma-efremoff, line 43: It looks like the end of the message will be duplicated by:
[`You could use the command bellow to load it locally`](https://github.com/ytsaurus/ytsaurus-ui/pull/709/commits/fc95d3a354448de118631ccd4f6e1d681fa83a62#diff-75d5eac1a43f5ae20c4ec63e1e668a32e26debc505e43464891c24efcefd8b34R53)

packages/ui/src/ui/containers/CellPreviewModal/CellPreviewModal.tsx
Flunt1k: Yes, it looks like it is better to replace it with noticeText and set the text itself in the action. Then we can put all the checks into action

ma-efremoff (approved):

Issue #712, opened 2024-08-08
Dark theme bug: unreadable description in navigation content tab


vrozaev:
<img width="491" alt="Screenshot 2024-08-08 at 10 38 10" src="https://github.com/user-attachments/assets/321938b0-0fba-4296-99f4-f9cb71482845">

<img width="475" alt="Screenshot 2024-08-08 at 10 38 00" src="https://github.com/user-attachments/assets/ea9cff28-230f-433e-af73-d4463ada786d">

PR #713, opened 2024-08-08
fix: added support for dark and light themes by override yfm styles [#712]

vrozaev:
Before:
image

After:
image


packages/ui/src/ui/entries/ytm-overrides.scss
```diff
@@ -0,0 +1,113 @@
+// Overrides for styles from @diplodoc/transform/dist/css/yfm.css
```
ma-efremoff, line 1: A misprint in the filename: `ytm-overrides.scss` -> `yfm-overrides.scss`. 
Is it works for you?

vrozaev: Fixed, sorry.

ma-efremoff (approved):

PR #716, opened 2024-08-12
feat: now we can setup uploadTableExcelBaseUrl and exportTableBaseUrl per cluster

vrozaev:
This is a draft PR which allows us to specify uploadTableExcelBaseUrl and exportTableBaseUrl per cluster.


packages/ui/src/shared/ui-settings.ts
```diff
@@ -9,8 +9,8 @@ export interface UISettings {
     uploadTableMaxSize: number;
     uploadTableUseLocalmode?: boolean;
 
-    uploadTableExcelBaseUrl?: string;
-    exportTableBaseUrl?: string;
+    uploadTableExcelBaseUrl?: string | Record<string, string>;
+    exportTableBaseUrl?: string | Record<string, string>;
```
ma-efremoff, line 13: Probably it is better to use `clusters-config.json` to provide cluster specific options, something like:
```ts
interface ClusterConfig {
    ...
    uiSettings?: Partial<Pick<UISettings, 'uploadTableExcelBaseUrl' | 'exportTableBaseurl'>>;
}
```

ma-efremoff: Then we can have simple function to get cluster-specific uiSettings, like:
```ts
function getClusterSpecificUISettings(clusterConfig?: ClusterConfig) {
    return {...uiSettings, ...clusterConfig.uiSettings};
}
```

packages/ui/src/ui/config/index.ts
```diff
@@ -3,6 +3,20 @@ import _ from 'lodash';
 import UIFactory from '../UIFactory';
 
 import {getConfigData, uiSettings} from './ui-settings';
+import YT from '../config/yt-config';
+import type {UISettings} from '../../shared/ui-settings';
+
+function getClusterSpecificUISettings({
+    uiSettings,
```
ma-efremoff, line 10: uiSettings is a constant value from `window.__DATA__`, so it is enough to have one parameter `cluster`.

ma-efremoff (approved):

PR #718, opened 2024-08-12
feat(Query): custom vcs [YTFRONT-4257]

SimbiozizV:
https://nda.ya.ru/t/oT2fclo_77Ykri

### DIFF
--- a/packages/ui/src/server/components/vcs/index.ts
+++ b/packages/ui/src/server/components/vcs/index.ts
@@ -1,39 +1,30 @@
 import {VCSSettings} from '../../../shared/ui-settings';
 import {GithubApi} from './GithubApi';
 import {GitlabApi} from './GitlabApi';
-import {AxiosResponse} from 'axios';
 import {Request} from '@gravity-ui/expresskit';
 import {AppConfig} from '@gravity-ui/nodekit';
 import {ErrorWithCode} from '../../utils';
-
-export interface VcsApi {
-    token: string;
-    api: string;
-    vcsId: string;
-
-    getRequestHeaders(): {Authorization: string};
-    getRepositories(): Promise<unknown>;
-    getBranches(projectId: string): Promise<string[]>;
-    getRepositoryContent(projectId: string, branch: string, path: string): Promise<unknown[]>;
-    getFileContent(
-        projectId: string,
-        branch: string,
-        path: string,
-    ): Promise<AxiosResponse<NodeJS.ReadableStream>>;
-}
+import ServerFactory from '../../ServerFactory';
+import {VCS_DEFAULT_MAX_FILE_SIZE} from '../../../shared/constants/vcs';
 
 export class VcsFactory {
     static createVcs(vcs: VCSSettings, token?: string) {
-        if (!token) throw new ErrorWithCode(400, 'Token is required');
-
-        const apiConfig = {token, api: vcs.api, vcsId: vcs.id};
+        const apiConfig = {
+            token,
+            baseUrl: vcs.baseUrl,
+            vcsId: vcs.id,
+            maxFileSize: vcs.maxFileSize || VCS_DEFAULT_MAX_FILE_SIZE,
+        };
         switch (vcs.type) {
             case 'github':
                 return new GithubApi(apiConfig);
             case 'gitlab':
                 return new GitlabApi(apiConfig);
-            default:
-                throw new ErrorWithCode(415, `Unsupported VCS type: ${vcs.type}`);
+            default: {
+                const vcsApi = ServerFactory.createCustomVcsApi(vcs.type, vcs, token);
+                if (!vcsApi) throw new ErrorWithCode(400, `Unsupported VCS type`);
+                return vcsApi;
+            }
         }
     }
 
--- a/packages/ui/src/server/controllers/vcs.ts
+++ b/packages/ui/src/server/controllers/vcs.ts
@@ -1,6 +1,6 @@
-import {AppRouteHandler, Response} from '@gravity-ui/expresskit';
+import {AppRouteHandler, Request, Response} from '@gravity-ui/expresskit';
 import axios from 'axios';
-import {VcsConfig, VcsRepository} from '../../shared/vcs';
+import {VcsRepository} from '../../shared/vcs';
 import {ErrorWithCode, sendError} from '../utils';
 import {VcsFactory} from '../components/vcs';
 
@@ -16,16 +16,18 @@ const ERROR_MESSAGE = {
     VCS_ID_REQUIRED: 'vcsId is required',
     VCS_NOT_SUPPORTED: 'The current vcsId is missing from the vcsSettings setting in UISettings',
     TOKEN_REQUIRED: 'Token is required',
+    TOKENS_ERROR: 'Error getting vcs tokens availability',
     PATH_REQUIRED: 'Path is required',
     BRANCH_REQUIRED: 'Branch is required',
     REPOSITORY_REQUIRED: 'Repository is required',
     UNAUTHORIZED:
         'The service cannot authorize you. Check the token. You can add new token in the section Settings -> VCS',
 };
+const VCS_COOKIE_PREFIX = 'vcs_token_';
 
 const cookieNameByVcsId = (vcsId?: string) => {
     if (!vcsId) throw new ErrorWithCode(400, ERROR_MESSAGE.VCS_ID_REQUIRED);
-    return `vcs_${vcsId}`;
+    return VCS_COOKIE_PREFIX + vcsId;
 };
 
 const sendApiError = (res: Response, error: unknown) => {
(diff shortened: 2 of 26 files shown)


packages/ui/src/shared/ui-settings.ts
```diff
@@ -146,7 +146,7 @@ export interface UISettings {
      * id - unique identifier of VCS
      * name - name of VCS. The name is displayed in the selector
      * api - URL to you VCS api
-     * type - 'gitlab' | 'github'
+     * type - 'gitlab' | 'github' | 'custom'
```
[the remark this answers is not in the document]
SimbiozizV: That would be a mistake. `type: 'github' | 'gitlab' | string; ' will always be `string`. I don't see a reason to make such a generalized type if we can make it concrete

packages/ui/src/server/components/vcs/index.ts
```diff
@@ -33,7 +26,7 @@ export class VcsFactory {
             case 'gitlab':
                 return new GitlabApi(apiConfig);
             default:
-                throw new ErrorWithCode(415, `Unsupported VCS type: ${vcs.type}`);
+                throw new ErrorWithCode(415, `Unsupported VCS type`);
```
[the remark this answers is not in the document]
SimbiozizV: > HTTP client error response code 415 Unsupported Media Type indicates that the server refuses to accept the request because the content format is not supported by the server.

I think 415 is more appropriate here

packages/ui/src/shared/ui-settings.ts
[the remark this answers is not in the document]
ma-efremoff: Also we can require some prefix like `custom_`:
```ts
interface VCSSettings {
    id: string;
    name: string;
    api: string;
    type: 'github' | 'gitlab' | CustomVCSType;
}

type CustomVCSType = `custom_${string}`;
```
The same restriction should be applyed to ServerFactory.createCustomVscApi:
```ts
interface ServerFactory {
    createCustomVscApi(type: CustomVCSType, ...): ...;
}
```

packages/ui/src/ui/pages/query-tracker/Vcs/Vcs.tsx
```diff
@@ -3,10 +3,26 @@ import {VcsHeader} from './VcsHeader';
...
+import {useSelector} from 'react-redux';
+import {selectConfigError} from '../module/vcs/selectors';
+import {Alert} from '@gravity-ui/uikit';
 
 const block = cn('vsc');
 
 export const Vcs: FC = () => {
+    const configError = useSelector(selectConfigError);
+
+    if (configError)
```
ma-efremoff, line 15: Does the error always mean `usSettings.vcsSettings` is not provided?

packages/ui/src/ui/pages/query-tracker/Vcs/VcsPath/VcsPathInputForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type Props = {
+    path: string;
+    onCancel: () => void;
+    onSave: (path: string) => void;
+};
+
+const block = cn('vcs-path-input-form');
+
+export const VcsPathInputForm: FC<Props> = ({path, onSave, onCancel}) => {
```
ma-efremoff, line 16: Usages of the component might be replaced by `components/EditableAsText`

packages/ui/src/ui/pages/query-tracker/Vcs/Vcs.tsx
```diff
@@ -3,10 +3,26 @@ import {VcsHeader} from './VcsHeader';
...
+        return (
+            <div className={block({error: true})}>
+                <Alert
+                    theme="info"
+                    title="vcsSettings is not found in UISettings"
+                    message="To work with vcs, you need to add vcsSettings in the cluster installation settings. Ask your administrator to add this setting"
```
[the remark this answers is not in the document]
SimbiozizV: The info mode is used here. We let the user know that this block is not working because it is not configured

packages/ui/src/ui/pages/query-tracker/Vcs/Vcs.tsx
SimbiozizV: This message is displayed if there are no vcs available in the settings at all. If there is at least 1 configuration, then there will be no block.

packages/ui/src/ui/pages/query-tracker/Vcs/VcsPath/VcsPathInputForm.tsx
SimbiozizV: This component doesn't fit the logic of this composition

packages/ui/src/shared/vcs.ts
```diff
@@ -1,21 +1,51 @@
...
+        repository: VcsRepository;
+        branch: string;
+        path: string;
+        requests?: Requests;
+    }): Promise<string>;
+}
+
+export interface VcsApi extends BaseVcsApi {
+    token: string;
+    baseUrl: string;
```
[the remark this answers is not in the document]
SimbiozizV: This is assumed to be the required field. A Vcs is a separate service that we need to interact with in some way. To do this, specify its address. 

For example, you can use regular GitHub and GitHub deployed somewhere on the internal network. In other words, the VCS type is the same, but the placements are different.

packages/ui/src/ui/pages/query-tracker/Vcs/Vcs.tsx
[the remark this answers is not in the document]
SimbiozizV: Снимок экрана 2024-08-30 в 12 25 27

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -22,10 +22,11 @@ const ERROR_MESSAGE = {
     UNAUTHORIZED:
         'The service cannot authorize you. Check the token. You can add new token in the section Settings -> VCS',
 };
+const VCS_COOKIE_PREFIX = 'vcs_';
```
ma-efremoff, line 25: `vcs_token_`

packages/ui/src/server/controllers/vcs.ts
```diff
@@ -50,24 +51,20 @@ const sendApiError = (res: Response, error: unknown) => {
...
     try {
-        const config = VcsFactory.getVcsSettings(req);
-        if (!config.length) throw new ErrorWithCode(404, ERROR_MESSAGE.VCS_CONFIG_NOT_FOUND);
-
-        const response = config.map<VcsConfig>((item) => ({
-            ...item,
-            hasToken: cookieNameByVcsId(item.id) in cookies,
-        }));
-        res.status(200).json(response);
+        const vcsTokenKeys = Object.keys(cookies).filter((name) =>
```
ma-efremoff, line 57: It is better to avoid sending of names of cookies with tokens to browser. 
It is enough to send id of such VCS.

packages/ui/src/ui/pages/query-tracker/module/vcs/vcsSlice.ts
```diff
@@ -44,19 +55,29 @@ const initialState: VcsState = {
...
     name: 'vcs',
     initialState,
     reducers: {
+        setVcsTokensAvailability(state, {payload}: PayloadAction<string[]>) {
+            const {vcsSettings} = uiSettings;
+            if (!vcsSettings) return;
+
+            state.vcsConfig = vcsSettings.map<VcsConfig>((item) => ({
+                ...item,
+                hasToken: payload.includes('vcs_' + item.id),
```
ma-efremoff, line 71: Browser code should know nothing about names of cookies with tokens.

packages/ui/src/ui/pages/query-tracker/Vcs/Vcs.tsx
```diff
@@ -3,10 +3,25 @@ import {VcsHeader} from './VcsHeader';
...
+import {Text} from '@gravity-ui/uikit';
+import {uiSettings} from '../../../config/ui-settings';
+import {Info} from '../../../components/Info/Info';
 
 const block = cn('vsc');
 
 export const Vcs: FC = () => {
+    const {vcsSettings} = uiSettings;
+
+    if (!vcsSettings)
```
ma-efremoff, line 15:
```ts
if (!vcsSettings?.length) {
    ...
}
```

packages/ui/src/shared/ui-settings.ts
```diff
@@ -185,11 +206,15 @@ export interface UISettings {
...
+
 export interface VCSSettings {
     id: string;
     name: string;
-    api: string;
-    type: 'github' | 'gitlab';
+    baseUrl: string;
+    auth: 'token' | 'none';
+    type: 'github' | 'gitlab' | CustomVCSType;
+    maxFileSize: number;
```
ma-efremoff, line 217: It doesn't look like a good idea to suppose megabytes for a field with name `maxFileSize`. Bytes is more appropriate here.

ma-efremoff (approved):

PR #735, opened 2024-08-21
fix(Navigation): error message in table [YTFRONT-4312]

SimbiozizV:
https://nda.ya.ru/t/U4ooiCFc77ijoC


ma-efremoff:
image
You have fixed only two places with `Promise.reject()`, but there are other ones. It is better to fix them too.

SimbiozizV:
Why should I process other Promise.Reject? The only place related to navigation is here:

Снимок экрана 2024-08-26 в 11 16 16

This function is used in one place, where there is an exception handler:

Снимок экрана 2024-08-26 в 11 18 41

SimbiozizV:
That's not so. Other places handle this kind of error. 
In 'schedulling', this is handled via `action` ' SCHEDULING_EDIT_POOL_FAILURE`.
In loadDynamicTable, this is handled at the call location. There are no problem areas where an error from the task might have occurred.

ma-efremoff (approved):

PR #751, opened 2024-09-02
fix(Navigation/MapNode): use 'navmode=auto' when node is clicked

Flunt1k:


ma-efremoff:
Let's provide more detailed title for your commit, like:
`fix(Navigation/MapNode): use 'navmode=auto' when node is clicked`

ma-efremoff (approved):

PR #752, opened 2024-09-02
feat(System): maintenance button [YTFRONT-4217]

SimbiozizV:
https://nda.ya.ru/t/CXx7rIvE77xvcb

### DIFF
--- a/packages/ui/src/ui/store/actions/system/masters.ts
+++ b/packages/ui/src/ui/store/actions/system/masters.ts
@@ -12,12 +12,18 @@ import {USE_SUPRESS_SYNC} from '../../../../shared/constants';
 import type {AxiosError} from 'axios';
 import type {Dispatch} from 'redux';
 import type {BatchSubRequest} from '../../../../shared/yt-types';
-import type {
+import {
     MasterAlert,
     MasterDataItemInfo,
     MastersConfigResponse,
+    MastersStateAction,
     ResponseItemsGroup,
 } from '../../reducers/system/masters';
+import {ThunkAction} from 'redux-thunk';
+import type {RootState} from '../../reducers';
+import {MasterInstance} from '../../selectors/system/masters';
+import {getMastersHostType} from '../../selectors/settings';
+import {VisibleHostType} from '../../../constants/system/masters';
 
 export const FETCH_MASTER_CONFIG = createActionTypes('MASTER_CONFIG');
 export const FETCH_MASTER_DATA = createActionTypes('MASTER_DATA');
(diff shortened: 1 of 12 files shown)


packages/ui/src/ui/utils/system/masters.ts
```diff
@@ -0,0 +1,13 @@
+import {makeRegexpFromSettings} from '../../../shared/utils';
+import {uiSettings} from '../../config/ui-settings';
+
+export const calcMasterShortName = (address: string) => {
```
ma-efremoff, line 4: The function looks like a copy-paste from `Host.tsx`, it is better to have only one version.

packages/ui/src/ui/utils/system/masters.ts
```diff
@@ -0,0 +1,13 @@
...
+    const re = makeRegexpFromSettings(uiSettings.reShortNameFromAddress);
```
ma-efremoff, line 5: It is better to think that creation of RegExp is no a cheap action and create an instance once.

packages/ui/src/ui/utils/system/masters.ts
```diff
@@ -0,0 +1,13 @@
...
+    if (re) {
+        const res = re?.exec(address);
+        if (res?.groups?.shortname) {
+            return [res.groups.shortname, res.groups.suffix].filter(Boolean).join('');
```
ma-efremoff, line 9: And by the way It is also might be replaced with more performant version, like:
```ts
return suffix ? `${shortname}${suffix}` : shortname;
```

packages/ui/src/ui/utils/system/masters.ts
```diff
@@ -0,0 +1,5 @@
+import {calcShortNameByRegExp} from '../../containers/Host/Host';
+
+export const calcMasterShortName = (address: string) => {
```
ma-efremoff, line 3: The function looks unnecessary, `address` value might be returned from `calcShortNameByRegExp`

packages/ui/src/ui/store/actions/system/schedulers.js
```diff
@@ -118,3 +122,34 @@ export function getAgents() {
...
+    ({address, maintenance, message, type}) =>
+    async (dispatch, getSate) => {
+        const state = getSate();
+        const isAgent = type === 'agents';
+        const items = isAgent ? getSystemAgents(state) : getSystemSchedulers(state);
+
+        const master = items.find((i) => {
+            const host = i.host.$value;
+            const annotations = ypath.getValue(i.host, '/@annotations');
+            return host === address || annotations.physical_host === address;
```
ma-efremoff, line 136: It is more strict to check only corresponding value according to current mode (`host`/`container`). Also it explicitly highlights dependencies in the code.

packages/ui/src/ui/store/actions/system/masters.ts
```diff
@@ -399,6 +403,96 @@ export function loadMasters() {
...
+        const instances: MasterInstance[] = [
+            ...primary.instances,
+            ...secondary.reduce<MasterInstance[]>((acc, item) => [...acc, ...item.instances], []),
+            ...queueAgents.instances,
+            ...providers.instances,
+        ];
+
+        const master = instances.find((i) => {
+            const {host, physicalHost} = i.toObject();
+            return host === address || physicalHost === address;
```
ma-efremoff, line 474: It is more strict to check only corresponding value according to current mode (`host`/`container`). Also it explicitly highlights dependencies in the code.

packages/ui/src/ui/store/actions/system/masters.ts
```diff
@@ -399,6 +403,96 @@ export function loadMasters() {
...
+        if (!path) throw new Error('Cant take path by master type');
+
+        const result = await changeMaintenance({
+            path: `${path}/${master.toObject().host}`,
+            login: state.global.login,
+            maintenance,
+            message,
+        });
+
+        const error = getBatchError(result as any, 'Failed to update master maintenance');
```
ma-efremoff, line 488: Do we really need `as any`  here?

PR #755, opened 2024-09-03
fix(Query): query page fixes [YTFRONT-4340]

SimbiozizV:


ma-efremoff:
`fix(Query): query page fixes [YTFRONT-4340]`
It is better to use names of page of the UI: `fix(Query): ...` -> `fix(Queries): ...`

PR #760, opened 2024-09-06
fix(Navigation/Consumer,Navigation/Queue): show errors [YTFRONT-4144]

ma-efremoff:
https://nda.ya.ru/t/8fjwGsK0784xyt


packages/ui/src/ui/components/WarningIcon/WarningIcon.tsx
```diff
@@ -11,14 +13,21 @@ const b = cn('yt-warning-icon');
...
 }
 
 export default function WarningIcon(props: Props) {
-    const {className, hoverContent} = props;
+    const {className, hoverContent, children, color} = props;
     return (
         <span className={b(null, className)}>
             <Tooltip className={b('tooltip')} content={hoverContent}>
-                <Icon className={b('icon')} awesome="exclamation-triangle" />
+                <Icon className={b('icon', {color})} awesome="exclamation-triangle" />
```
SimbiozizV, line 25: We don't want to slowly abandon the old components? Maybe we should replace it with an `Icon` from data-ui with this [icon](https://gravity-ui.com/icons?icon=TriangleExclamation).

packages/ui/src/ui/pages/navigation/tabs/Queue/utils/column-builder.tsx
```diff
@@ -37,19 +39,40 @@ export function datetime<T>(name: string, getter: (row: T) => string | null): Co
     };
 }
 
-export function error<T>(name: string, getter: (row: T) => YTError | undefined): Column<T> {
+export function error<T>(
+    name: string,
+    getter: (row: T) => YTError | undefined,
+    {className}: {className?: string} = {},
```
SimbiozizV, line 45: Are you sure you need an object here? I don't see any other parameters other than `className`. Are there any other parameters planned in the future? 

I offer 2 possible solutions:
1) Move all attributes to a single object: `{name, getter, className}: {name: string,.....}`, so you can flexibly add new parameters without editing places where this function is already used.
2) Extract className from the object: `(name: string, ...., className?: string)`. If no new parameters are planned, then we don't need the object there either.

packages/ui/src/ui/store/selectors/operations/operation.ts
```diff
@@ -31,7 +33,41 @@ interface OperationTask {
...
+                appeared[type] = acc[acc.length - 1];
+            } else {
+                acc.push({
+                    to: ypath.getValue(item.time),
+                    type,
+                    error: item.error,
+                    url: prepareFaqUrl(type),
+                });
+            }
+            return acc;
```
SimbiozizV, line 66: I would not recommend the `if {} else if {} else {}` construction here. It complicates the readability of the code. I would just do `return acc` after the logical block:
if(code) {...; return acc;}
if(appeared[type]) {...; return acc;}
.... return acc;

This is not a mistake, so it's up to you.

SimbiozizV (approved):

PR #767, opened 2024-09-12
feat(MapNode): added posibility so setup custom icons and load additional attributes for map-node

vrozaev:


packages/ui/src/ui/components/Icon/importGravityIcons.ts
```diff
@@ -15,4 +15,5 @@ export const iconNames = {
     ['layout-side-content']: LayoutSideContent,
     ['sql']: AbbrSql,
     ['crowndiamond']: CrownDiamond,
+    ['python']: LogoPython,
```
ma-efremoff, line 18: It is never used directrly. Do we really need it?

packages/ui/src/ui/pages/navigation/content/MapNode/MapNodesTableIcon.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Icon from '../../../../components/Icon/Icon';
+import {getIconNameForType} from '../../../../utils/navigation/path-editor';
+
+const TRASH_PATH = '//tmp/trash';
+
+export function isTrashNode(item: any) {
```
ma-efremoff, line 7: It is always bad idea to use `any` this way. Required type might be easily inferred from implementation:
```ts
export function isTrashNode(item: {path: string}) {
    ...
}
```

packages/ui/src/ui/pages/navigation/content/MapNode/MapNodesTableIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+    return item.path === TRASH_PATH;
+}
+
+export function isLinkToTrashNode(item: any) {
```
ma-efremoff, line 11: `any`!

packages/ui/src/ui/pages/navigation/content/MapNode/MapNodesTableIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+    return item.targetPath === TRASH_PATH;
+}
+
+export function renderMapNodesTableIcon(item: any) {
```
ma-efremoff, line 15: `any`!
Let's keep the method on its previous place. And add call of corresponding `UIFacotry`-method, like:
```ts
export function renderMapNodesTableIcon(item) {
    let icon = UIFactory.getNavigationMapNodeSettings()?.renderNodeIcon(item);
    if (icon) {
       // do nothing
    } else if ...
}
```

packages/ui/src/ui/components/Icon/importGravityIcons.ts
vrozaev: You are right, it's not neccesery, I will roll it out.

packages/ui/src/ui/UIFactory.tsx
```diff
@@ -426,6 +427,11 @@ export interface UIFactory {
         menuItems: Array<DropdownMenuItem>;
         renderModals: () => React.ReactNode;
     };
+
+    getMapNodeSettings(): {
+        additionalAttributes: unknown[];
+        renderMapNodesTableIcon: (item: unknown) => React.ReactNode;
```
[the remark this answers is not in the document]
vrozaev: I will move this file packages/ui/src/ui/utils/navigation/content/map-nodes/node.js to typescript, that will give us accurate types.

ma-efremoff (approved):

PR #769, opened 2024-09-17
fix(YQLTable): fixed exception in truncated cells preview

vrozaev:
Error in interface
image

Error in debug mode
Screenshot 2024-09-17 at 10 12 20

The root case - we don't have an array in `response.rows[0][options.columnName][0]`

Screenshot 2024-09-17 at 10 12 28

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/module/cell-preview/actions.ts
+++ b/packages/ui/src/ui/pages/query-tracker/module/cell-preview/actions.ts
@@ -27,7 +27,13 @@ export const showQueryTrackerCellPreviewModal = (
                 ),
             );
 
-            const data = (response.rows[0][options.columnName][0] as any[])[0].val;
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/pages/query-tracker/module/cell-preview/actions.ts
```diff
@@ -27,7 +27,9 @@ export const showQueryTrackerCellPreviewModal = (
                 ),
             );
 
-            const data = (response.rows[0][options.columnName][0] as any[])[0].val;
+            const dataRow = response.rows[0][options.columnName][0] as any | any[];
+
+            const data = Array.isArray(dataRow) ? dataRow[0].val : dataRow.val;
```
SimbiozizV, line 32: My concern is `dataRow[0].val`. Won't there be a situation where the array will come empty?

SimbiozizV (approved):

PR #789, opened 2024-09-30
fix(Query): do not show vcs if vcsSettings is empty

vrozaev:


packages/ui/src/ui/pages/query-tracker/module/vcs/vcsSlice.ts
```diff
@@ -34,15 +35,24 @@ export type VcsState = {
...
 
-    return vcsSettings.map((item) => ({...item, hasToken: false}));
+    if (!vcsSettings) {
+        return {
+            hasVcsSettings: false,
+            vcsConfig: [],
+        };
+    }
+
+    return {
```
SimbiozizV, line 48: I don't like that you're introducing a separate `hasVcsSettings` flag here. I'm not sure it's needed here.

I would change the type of the `vcsConfig` variable and explicitly specify undefined if it is not set in the config. So it would be possible to get 3 states (not set, want to hide, set). This way it will be a little more correct and there will not be an extra variable.

vrozaev: Sure, will change it

vrozaev: Done

SimbiozizV (approved):

PR #800, opened 2024-10-09
feat(System): new regexp shortname [YTFRONT-4386]

SimbiozizV:
https://nda.ya.ru/t/NzV8RS2z78oBSH

### DIFF
--- a/packages/custom-ytsaurus-ui.example/src/server/configs/common.ts
+++ b/packages/custom-ytsaurus-ui.example/src/server/configs/common.ts
@@ -21,6 +21,7 @@ const cfg: Partial<CustomAppConfig> = {
         reHashFromNodeVersion: '[^~]+~(?<hash>[^+]+)',
 
         reShortNameFromAddress: '(?<shortname>.*)(\\.[^.]+)(\\.yt\\.my\\.domain)',
+        reShortNameSystemPage: '(?<shortname>.*)(\\.[^.]+)(\\.yt\\.my\\.domain)',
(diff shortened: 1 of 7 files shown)


packages/custom-ytsaurus-ui.example/src/server/configs/common.ts
```diff
@@ -21,6 +21,7 @@ const cfg: Partial<CustomAppConfig> = {
         reHashFromNodeVersion: '[^~]+~(?<hash>[^+]+)',
 
         reShortNameFromAddress: '(?<shortname>.*)(\\.[^.]+)(\\.yt\\.my\\.domain)',
+        reShortSystemAddressName: '(?<shortname>.*)(\\.[^.]+)(\\.yt\\.my\\.domain)',
```
ma-efremoff, line 24: Isn't it better to use similar names for options?
`reShortNameFromMasterAddress`/`reShortNameSystemPage`

ma-efremoff: As I remember `reShortNameFromAddress` is covered with a test. Is it possible to add similar one for the new option?

SimbiozizV: I added it to the e2e configuration by analogy with the other 2.
But I don't see the point in covering a regular expression that changes in the configuration with a test.

ma-efremoff (approved):

PR #811, opened 2024-10-14
some fixes

ma-efremoff:
https://nda.ya.ru/t/SLXSbjNQ78vJvk


packages/ui/src/ui/pages/components/tabs/nodes/tables.tsx
```diff
@@ -468,9 +513,24 @@ const nodesTableProps = {
...
+                                    column: 'memory_usage',
+                                    title: 'Usage',
+                                    withUndefined: true,
+                                    allowUnordered: true,
+                                },
+                                {
+                                    column: 'memory_limit',
+                                    title: 'Limit',
+                                    withUndefined: true,
+                                    allowUnordered: true,
```
SimbiozizV, line 532: I don't really like that we have a lot of blocks with duplicate parameters `withUndefined: true, allowUnordered: true`

I suggest putting them in a constant and adding them via `{title: '...', ...MY_CONST}`

SimbiozizV (approved):

Issue #841, opened 2024-10-28
Support multiply pool tree selection for strawberry (chyt) creation dialog


vrozaev:
Look like we need to change [PoolTreeSuggestControl](https://github.com/ytsaurus/ytsaurus-ui/blob/22879ac538c249ef6c4ff67c415aa49ea6f840dc/packages/ui/src/ui/components/Dialog/controls/PoolTreeSuggestControl/PoolTreeSuggestControl.tsx#L14) and review it's usages in [PoolSuggestControl](https://github.com/ytsaurus/ytsaurus-ui/blob/22879ac538c249ef6c4ff67c415aa49ea6f840dc/packages/ui/src/ui/components/Dialog/controls/PoolSuggestControl/PoolSuggestControl.tsx#L45) component and other places.

Right now we use PoolTreeSuggestControl in 3 places, all of them should be checked:

1. Clique creation - we should be able to select several pool_trees here.
<img width="730" alt="Screenshot 2024-10-28 at 14 51 34" src="https://github.com/user-attachments/assets/913c40fd-46d2-487c-ae9e-0f72cbab04da">

2. Table merge - nothing should change
<img width="723" alt="Screenshot 2024-10-28 at 14 57 18" src="https://github.com/user-attachments/assets/0f9af0e5-6ed1-4072-a7d2-1610e9b47346">

3. Table Sort - nothing should change
<img width="724" alt="image" src="https://github.com/user-attachments/assets/59192543-61d7-426d-892c-11f1508680ab">

PR #842, opened 2024-10-30
feature: support multiply pool tree selection

vrozaev:
https://nda.ya.ru/t/uLQGj02C79Trsr
    
https://github.com/ytsaurus/ytsaurus-ui/issues/841


packages/ui/src/ui/components/Dialog/df-dialog-utils.ts
```diff
@@ -116,6 +117,8 @@ export function descriptionToDialogField<T = unknown>(
             return {...common, type: 'path', extras};
         case 'pool':
             return {...common, type: 'pool', extras: {...extras, allowEmpty: true}};
```
vrozaev, line 119: I'm also want to pass here poolTrees value somehow, but I'm not sure that it is right place. Without poolTrees in extra the suggest will not work properly (will work only with default pool tree).

```
(form) => {
        return {
            poolTrees: form.group_0.pool_trees,
        };
    }
```

ma-efremoff (approved):

PR #847, opened 2024-11-06
feat(Queries): allow users to copy query id

vitshev:
https://nda.ya.ru/t/tekwD2Od79Wrpr


packages/ui/src/ui/pages/query-tracker/QueryMetaTable/index.tsx
```diff
@@ -21,13 +22,16 @@ const valueClassName = b('value');
...
     value: string | number | boolean;
+    showCopy: boolean;
 };
 
 const exceptFields: (keyof QueryItem | 'access_control_object')[] = [
     'query',
     'access_control_object',
 ];
 
+const FIELDS_WITH_COPY_BUTTON = ['id'];
```
SimbiozizV, line 33: Is it planned that there will be more such fields? Now it would be easier to use string.

PR #855, opened 2024-11-12
fix(System): timestamp providers maintenance [YTFRONT-4452]

SimbiozizV:
https://nda.ya.ru/t/KQtneYyl79baPA


packages/ui/src/ui/pages/system/Masters/Instance.tsx
```diff
@@ -88,10 +98,10 @@ export const Instance: FC<Props> = ({instance, hostType, allowVoting, allowServi
                     {allowService && (
                         <ChangeMaintenanceButton
                             className={b('host-btn')}
-                            address={address}
+                            path={`${getPathByMasterType(instance.getType())}/${instance.getHost()}`}
```
ma-efremoff, line 101: Can we add `MasterInstance.getPath()` and call it here?

packages/ui/src/ui/pages/system/Masters/Instance.tsx
```diff
@@ -12,110 +11,96 @@ import ClipboardButton from '../../../components/ClipboardButton/ClipboardButton
...
-        // do not use `!voting` cause `voting === undefined` is the same as `voting === true`
-        const denyVoting = allowVoting && voting === false;
-        const theme =
-            denyVoting && state === 'following'
-                ? 'nonvoting'
-                : Instance.instanceStateToTheme[state];
-        const addressWithoutPort = hammer.format['Address'](address);
+export const Instance: FC<Props> = ({instance, hostType, allowVoting, allowService}) => {
+    const {state, $address, $physicalAddress} = instance;
+    const attributes = instance.$attributes as Record<string, any>;
```
ma-efremoff, line 46: Is it possible to provide more details for the type (instead of `Record<string, any>`)?
(also we have helper type `CypressNode`)

packages/ui/src/ui/pages/system/Masters/ChangeMaintenanceButton.tsx
```diff
@@ -1,55 +1,52 @@
...
 import PencilIcon from '@gravity-ui/icons/svgs/pencil.svg';
 import {FormApi, YTDFDialog, makeErrorFields} from '../../../components/Dialog';
-import hammer from '../../../common/hammer';
 import {useToggle} from 'react-use';
 
 type Props = {
     className?: string;
-    address: string;
+    path: string | null;
+    title?: string;
```
ma-efremoff, line 10: The title might be always generated from `path`:
```ts
const address = path?.split('/').pop();
const title = `Edit ${address}`;
```
Do we really need the property?

packages/ui/src/ui/pages/system/Masters/Instance.tsx
```diff
@@ -88,10 +102,12 @@ export const Instance: FC<Props> = ({instance, hostType, allowVoting, allowServi
                     {allowService && (
                         <ChangeMaintenanceButton
                             className={b('host-btn')}
-                            address={address}
+                            path={instance.getPath()}
+                            title={`Edit ${shortName}`}
```
ma-efremoff, line 106: `shortName` doesn't look like a good choice here. It is better to show full address (not path).

packages/ui/src/ui/pages/system/Masters/ChangeMaintenanceButton.tsx
SimbiozizV: The header is formed differently for Scheduler and Masters. That's why I made it even higher

evtaranov (approved):

PR #859, opened 2024-11-12
fix(Queries): run query hotkeys [YTFRONT-4462]

SimbiozizV:
https://nda.ya.ru/t/XtNnR1O-79bvfR


packages/ui/src/ui/pages/query-tracker/QueryEditor/QueryEditor.tsx
```diff
@@ -100,9 +100,9 @@ const QueryEditorView = React.memo(function QueryEditorView({
             }
         };
 
-        document.addEventListener('keydown', runQueryByKey);
+        document.addEventListener('keydown', runQueryByKey, true);
```
ma-efremoff, line 103: Should't we use `hotkeys-js` here as in other places of our project?

ma-efremoff (approved):

Issue #634, opened 2024-06-20
Management of groups through UI


zlobober:
This issue is similar to #633, but for the groups.

There is a "/groups" page displaying all groups.

1) Let's add a yellow button for creating a group at the top right corner

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/c5ec9d08-b010-494e-bf92-28ec6114291a)

Modal should ask only for the group name. To create a group, the `create("group", attributes={"name": "<name>"})` should be issued.

2) Let's add an "@" icon opening the attributes of a group, similar to how it is done on the Users page:

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/0282c00e-1ba2-4a94-80f4-879bf18b3568)
![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/adc518ae-1e0c-4175-9031-d469550d2e2f)
![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/6fd33210-cad7-467e-82ac-fa8b8d1d9ca6)

3) Let's add a pen icon opening the modal window for editing the group, similar to the users interface

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/c5759b9f-f79c-47a4-bf8c-1ed3dd8a57d7)

![image](https://github.com/ytsaurus/ytsaurus-ui/assets/1856442/1952549f-9e8c-4f7a-9814-940107d1c2be)

3.1) There should be a "General" tab, on which there should be the only field "group name". In order to change the group name, a Cypress request set("//sys/groups/<group>/@name", "<new group>") should be issued.

3.2) On another tab "members" I suggest to have a standard two column interface for adding/removing members from a group.

```
 Group members      |     All users and groups
 [search]           |         [search]
 alice              |         charlie
 bob                <-        john
 joe                ->        admins    
 devs               |         analytics
                    |                   ...


[Save]
```

(ChatGPT suggests that this thing is called a "dual list box")

Left column should consist of `//sys/groups/<group>/@members`. Right column should consist of "everybody else", which is the union of `list("//sys/users")` and `list("//sys/groups")` minus the members of the group. Both lists should be in the alphabet order.

A "search" should perform in-memory filtration by a substring. 

Press of "save" should issue a batch of `add_member("<member>", "<current group>")` or `remove_member("<member>, "<current group>")` which remove the and add the members that were included or excluded from the original list.

[...]

PR #863, opened 2024-11-14
chore(Groups): rewrite group modules to typescipt [#634]

vrozaev:


packages/ui/src/ui/store/selectors/groups.ts
```diff
@@ -127,33 +140,34 @@ export const getGroupsFlattenTree = createSelector([getGroupsTreeFilteredAndSort
...
+const getGroupEditorIdmData = (state: RootState) => state.groups.editor.idmData;
+export const getGroupEditorIdmDataVersion = (state: RootState) =>
+    state.groups.editor.idmData.version;
+export const getGroupEditorIdmDataOtherMembers = (state: RootState) =>
     state.groups.editor.idmData.group.other_members;
 
 export const getGroupEditorSubjects = createSelector([getGroupEditorIdmData], (idmData) => {
     const {
-        group: {members, responsible},
+        group: {members, responsibles},
```
vrozaev, line 157: That typo was here all along. 

I suspect that idm and group editor were never tested in opensource yt.

packages/ui/src/ui/constants/groups.ts
```diff
@@ -0,0 +1,9 @@
+import createActionTypes from '../constants/utils';
```
vrozaev, line 1: I manually renamed these files via `git mv`, but once again - git decided that it's a new file.

packages/ui/src/ui/store/reducers/groups/table.ts
```diff
@@ -0,0 +1,105 @@
+import {GROUPS_TABLE, GROUPS_TABLE_DATA_FIELDS} from '../../../constants/groups';
```
vrozaev, line 1: I manually renamed these files via `git mv`, but once again - git decided that it's a new file.

packages/ui/src/ui/store/selectors/groups.ts
```diff
@@ -61,10 +73,10 @@ const getGroupsTreeFiltered = createSelector(
         const root = tree[ROOT_GROUP_NAME];
         const predicates = [
             nameFilter &&
-                ((node) => {
-                    return node === root || -1 !== node.name.indexOf(nameFilter);
+                ((node: GroupsTreeNode) => {
+                    return node === root || -1 !== node.name!.indexOf(nameFilter);
                 }),
-        ].filter(Boolean);
+        ].filter(Boolean) as Array<(node: GroupsTreeNode) => boolean>;
```
ma-efremoff, line 80: It is better to use `compact` from `lodash` instead of `.filter(Boolean)`, cause it does good inferring of types and you won't need to add expressions like `as ...`.

ma-efremoff (approved):

Issue #853, opened 2024-11-08
Show all groups in UI


Krock21:
Currently, inner groups are not shown on a top level. This breaks user scenario "Quickly filter all groups and check their members"

For example, here I just wanted to check "superusers" group, but was shown "everyone" instead, and had to click 2 times to unfold it
<img width="1511" alt="Screenshot 2024-11-08 at 18 51 27" src="https://github.com/user-attachments/assets/c0a7c2dd-a1bf-4a63-8953-b1c3176e2c1b">
<img width="1511" alt="Screenshot 2024-11-08 at 18 51 40" src="https://github.com/user-attachments/assets/43269a32-3825-40db-b8d4-c755ae7ada96">

PR #887, opened 2024-11-26
fix(Groups): expand group tree on non-empty group filter [#853]

vrozaev:
Let's say we have 3 groups:

- vrozaev-test-group-1
- vrozaev-test-group-2
- vrozaev-test-group-3

And the hierarchy is: vrozaev-test-group-1 => vrozaev-test-group-2 => vrozaev-test-group-3

In this case we want to expand all parents of the node which is matches the filter.

1. Example: with "vrozaev-test-group-1" filter there will be no expanded nodes
Screenshot 2024-11-26 at 11 26 11

1. Example: with "vrozaev-test-group-2" filter the "vrozaev-test-group-1" node will be expanded
Screenshot 2024-11-26 at 11 26 19

1. Example: with "vrozaev-test-group-3" filter the "vrozaev-test-group-1" and "vrozaev-test-group-2" nodes will be expanded
Screenshot 2024-11-26 at 11 26 25

### DIFF
--- a/packages/ui/src/ui/store/selectors/groups.ts
+++ b/packages/ui/src/ui/store/selectors/groups.ts
@@ -88,15 +88,38 @@ const getGroupsTreeFiltered = createSelector(
 );
 
 const getGroupsTreeFilteredAndExpanded = createSelector(
-    [getGroupsTreeFiltered, getGroupsExpanded],
-    (root, expanded) => {
-        const res = cloneDeep_(root);
+    [getGroupsTree, getGroupsTreeFiltered, getGroupsExpanded, getGroupsNameFilter],
+    (groupTree, groupsTreeFiltered, expandedByUser, groupNameFilter) => {
+        const expandedBySearch: Record<string, boolean> = {};
+
+        const res = cloneDeep_(groupsTreeFiltered);
+
         hammer.treeList.treeForEach(res.children, (node: GroupsTreeNode) => {
-            const {name} = node;
-            if (!expanded[name]) {
-                node.children = [];
+            const isNodeMatchedFilter = groupNameFilter && node.name.includes(groupNameFilter);
+
+            if (isNodeMatchedFilter) {
+                let parentName = node.parent!;
+
+                while (groupTree[parentName]) {
+                    expandedBySearch[parentName] = true;
+                    parentName = groupTree[parentName].parent!;
+                }
             }
         });
+
+        hammer.treeList.treeForEach(res.children, (node: GroupsTreeNode) => {
+            const userInteractedWithNode = typeof expandedByUser[node.name] !== 'undefined';
+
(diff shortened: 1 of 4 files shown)


packages/ui/src/ui/store/selectors/groups.ts
```diff
@@ -88,15 +88,33 @@ const getGroupsTreeFiltered = createSelector(
...
+        const res = cloneDeep_(groupsTreeFiltered);
+
+        hammer.treeList.treeForEach(res.children, (node: GroupsTreeNode) => {
+            const isNodeMatchedFilter = groupNameFilter && node.name.includes(groupNameFilter);
+
+            if (isNodeMatchedFilter) {
+                let parentName = node.parent!;
+
+                while (groupTree[parentName]) {
+                    expanded[parentName] = true;
```
ma-efremoff, line 106: The changes enforce expanding of parents. What should user do if he wants to collapse an unnecessary branch?

vrozaev: Fixed, now user can do that

packages/ui/src/ui/store/selectors/groups.ts
```diff
@@ -88,15 +88,38 @@ const getGroupsTreeFiltered = createSelector(
...
-            const {name} = node;
-            if (!expanded[name]) {
+            const isNodeMatchedFilter = groupNameFilter && node.name.includes(groupNameFilter);
+
+            if (isNodeMatchedFilter) {
+                let parentName = node.parent!;
+
+                while (groupTree[parentName]) {
+                    expandedBySearch[parentName] = true;
+                    parentName = groupTree[parentName].parent;
```
ma-efremoff, line 105: Our internal version of UI builds with the error:
```
ERROR in ./src/ytsaurus-ui.ui/store/selectors/groups.ts:[32m[1m105:21[22m[39m
TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
    103 |                 while (groupTree[parentName]) {
    104 |                     expandedBySearch[parentName] = true;
  > 105 |                     parentName = groupTree[parentName].parent;
        |                     ^^^^^^^^^^
    106 |                 }
    107 |             }
    108 |         });
```
Something wrong with build of os-version of UI, it doesn't catch the error, we need to fix it someday.
Please add `!` character at the end of the line:
```ts
parentName = groupTree[parentName].parent!;
```

ma-efremoff (commented):
It looks like we need to click twice to collapse a group expaded by filter. Is it works the same for you?

vrozaev:
> It looks like we need to click twice to collapse a group expaded by filter. Is it works the same for you?

Yes you are right. In order to fix that I will change ExpandIcon component and toggleGroupExpand function

PR #894, opened 2024-11-28
feat(ManageTokens): optional password for token issuing

vitshev:
We want to enable SSO for our clusters and after that users wouldn’t have passwords, but they still need to issue tokens. 

Related backend issue https://github.com/ytsaurus/ytsaurus/issues/927


packages/ui/src/ui/containers/ManageTokens/ManageTokensModalContent/ManageTokensModalContent.tsx
```diff
@@ -393,6 +414,7 @@ export const ManageTokensModalContent = () => {
             case ViewSection.tokens:
                 return (
                     <AuthenticationTokensSection
+                        loadTokens={isOAuth}
```
ma-efremoff, line 417: It looks like we will not load tokens for `authWay === 'passwd'`, will we?

ma-efremoff (approved):

ma-efremoff (approved):

PR #895, opened 2024-11-28
some fixes

ma-efremoff:


packages/ui/src/ui/store/actions/acl.ts
```diff
@@ -59,20 +57,10 @@ function getPathToCheckPermissions(
...
-                return `//sys/pool_trees/${poolTree}/${poolPath.join('/')}`;
-            } else {
-                const params = new URLSearchParams(window.location.search);
-                const path = params.get('path');
-                if (!path) {
-                    break;
-                }
-                return path;
-            }
+            const fullPath: string = await ytApiV3Id.get(YTApiId.schedulingPoolFullPath, {
```
SimbiozizV, line 60: error handling?

packages/ui/src/ui/common/thor/getNumber.spec.js
```diff
@@ -65,5 +69,13 @@ describe('./thor/utils', () => {
                 expect(getNumber(node, '/@test_null', NaN)).toBe(NaN);
             });
         });
+
+        describe('yson', () => {
+            it('test', () => {
```
SimbiozizV, line 74: the name doesn't show the test goal

packages/ui/src/ui/store/actions/acl.ts
ma-efremoff: It is by design, the code should throw error.

packages/ui/src/ui/common/thor/getNumber.spec.js
SimbiozizV: number contert test?

SimbiozizV (approved):

PR #898, opened 2024-12-02
fix(Operations): job tooltip value format [YTFRONT-4211]

SimbiozizV:
https://nda.ya.ru/t/6XtKoHtb7A6vXi


evtaranov (approved):
Looks good

PR #920, opened 2024-12-16
fix(OAuth): fix redirect to previous page instead of /

Flunt1k:


packages/ui/src/@types/core.d.ts
```diff
@@ -110,6 +110,10 @@ export interface YTCoreConfig {
          *  Label on the Login via OpenID button
          */
         buttonLabel?: string;
+        /**
+         * Represents the URL to which request will be redirected.
+         */
+        redirectUrl?: string;
```
ma-efremoff, line 116: callbackBaseUrl/callBackOrigin

packages/ui/src/server/components/oauth.ts
```diff
@@ -72,14 +78,21 @@ export function saveOAuthTokensInCookies(res: Response, tokens: OAuthAuthorizati
     }
 }
 
-export function getOAuthLoginPath(req: Request) {
+export function getOAuthLoginPath(req: Request, res: Response) {
+    const {origin, pathname, search} = new URL(req.get('referrer') ?? getRedirectURL(req));
```
ma-efremoff, line 82: The value from config should have higher priority than origin from `referrer`

ma-efremoff (approved):

PR #922, opened 2024-12-17
feat(Login): add ability to customize login logo

vrozaev:


vrozaev:
Discussed on SIG UI meeting, will not be done because it is strange to provide such possibility - we want keep & develop ytsaurus brand.

PR #947, opened 2025-01-16
Make job abortion and abandonding warnings more precise

gritukan:


vrozaev:
That PR will be closed, here is the new one instead: https://github.com/ytsaurus/ytsaurus-ui/pull/948

PR #956, opened 2025-01-23
tmp

ma-efremoff:
tmp


packages/ui/src/ui/containers/MaintenancePage/HandleMaintenance.tsx
```diff
@@ -1,51 +1,37 @@
...
-                        }}
-                    />
-                );
-            }
+export function HandleMaintenance({cluster, children, emptyMaintenance}: Props) {
+    const [c, forceUpdate] = React.useState(0);
+
+    const maintenancePageEvent = useSelector(getMaintenanceEvent);
+
+    if (maintenancePageEvent) {
```
SimbiozizV, line 19: I would add an early return from the function.
This is not a mistake but a recommendation

```
if (! maintenancePageEvent) {
    return children;
}

if (emptyMaintenance) {
   return null;
}

return <MaintenancePage
                    cluster={cluster}
                    maintenancePageEvent={maintenancePageEvent}
                    onProceed={() => {
                        setMaintenanceIgnored(cluster);
                        forceUpdate(c + 1);
                    }}
                />
```

SimbiozizV (approved):

Issue #938, opened 2025-01-09
UI should display an error when user has no permission to set ACL


vitshev:
I don't see any errors in UI when I try to change acl without having enough permissions, only in the browser's console.

vrozaev:
PR with fix: https://github.com/ytsaurus/ytsaurus-ui/pull/962

PR #962, opened 2025-01-28
fix(ACL): UI should display an error when user has no permission to set ACL [#938]

vitshev:


packages/ui/src/ui/utils/acl/acl-api.ts
```diff
@@ -246,7 +248,17 @@ export function requestPermissions(params: RequestPermissionParams): Promise<Upd
...
+
         if (error) {
+            toaster.add({
+                name: 'get operation',
+                autoHiding: false,
+                theme: 'danger',
+                title: 'Failed to request permissions',
+                content: error.message,
+                actions: [{label: ' view', onClick: () => showErrorPopup(error)}],
+            });
```
ma-efremoff, line 260: We prefer to display error as close to its data as possible. Why aren't such errors displayed in their dialog?

ma-efremoff (approved):

vrozaev:
PR with fix: https://github.com/ytsaurus/ytsaurus-ui/pull/962

Issue #937, opened 2025-01-06
Deletion of nodes should be more intuitive


Krock21:
### SIG UI suggestion:

We discussed an issue and decide that following changes should be sufficient to solve the problem:

- [x] On action button, change text from "Delete" to "Move to trash" in case if "Delete permanently" checkbox disabled 
- [x] On modal window, change title from "Delete" to "Move to trash" in case if "Delete permanently" checkbox disabled 
- [x] Change title of success notification from "Objects deleted" to "Objects moved"
- [ ] In the success notification "Objects moved", show directory path where objects were moved, so users can go there and see them/undo

### Original issue decsriptiuon

Currently, deleting nodes in UI shows:
- Delete permanently checkbox, with no hint why is it different from "Delete"
- Delete button, which is actually a "Move to trash" button
- Confirmation notification with "Objects deleted" title and "Objects have been moved to the trash" description

<img width="649" alt="Screenshot 2025-01-06 at 12 10 39" src="https://github.com/user-attachments/assets/8c2c142a-4813-4bef-aba8-c327717e9838" />
<img width="312" alt="Screenshot 2025-01-06 at 12 11 01" src="https://github.com/user-attachments/assets/b8bec108-d437-4f05-84df-a7a06599c7b4" />

Some users assume that clicking "Delete" deletes nodes, and then notice resource leaks in their accounts

We should make a clear distinction between "Delete" and "Move to trash", and show a path for the trash in the notification

So I suggest:
1. Remove "Delete permanently" checkbox
2. Instead of "Delete" button, show 2 buttons:
   - Move to trash
   - Delete permanently
3. In the notification for "Move to trash", show trash directory path, so users can go there and see/undo

vrozaev:
> In the success notification "Objects moved", show directory path where objects were moved, so users can go there and see them/undo

We do have a problem here: objects could move to **different** directories, here is a function which decide where object should be moved:

```
// account - object account
// name - object name
// login - login of user which decided to delete object
const createDestinationPath = (account: string, name: string, login: string) =>
    `//tmp/trash/by-account/${account}/${login}/${name}_${moment().unix()}`;
```

That function works well with 1 object and we do show a link to trash in that case, but obviously we can't show any link in case when we trying to remove several objects.

vrozaev:
PR with fix: https://github.com/ytsaurus/ytsaurus-ui/pull/965

PR #965, opened 2025-01-29
fix(DeleteObjectModal): show different texts in case when user trying to remove objects permanently

vrozaev:
https://nda.ya.ru/t/e9ewEmQ37BYxCo
 Issue: https://github.com/ytsaurus/ytsaurus-ui/issues/937

### SIG UI suggestion:

We discussed an issue and decide that following changes should be sufficient to solve the problem:

- [x] On action button, change text from "Delete" to "Move to trash" in case if "Delete permanently" checkbox disabled 
- [x] On modal window, change title from "Delete" to "Move to trash" in case if "Delete permanently" checkbox disabled 
- [x] Change title of success notification from "Objects deleted" to "Objects moved"
- [x] In the success notification "Objects moved", show directory path where objects were moved, so users can go there and see them/undo


packages/ui/src/ui/pages/navigation/content/Table/TableOverview/EditTableActions.tsx
```diff
@@ -23,6 +23,7 @@ export default function EditTableActions() {
             dropDownBtnTheme={'outlined'}
             dropDownBtnSize={'m'}
             item={{
+                name: '',
```
vrozaev, line 26: Intentional code change during rewriting to typescript, looks safe for me.

packages/ui/src/ui/store/reducers/navigation/modals/delete-object.ts
```diff
@@ -19,15 +65,18 @@ export const initialState = {
...
+    resourceUsage: {
+        disk_space: 0,
+        node_count: 0,
+    },
     visible: false,
     permanently: false,
     multipleMode: false,
     multipleInfo: [],
-    item: {},
+    item: [],
```
vrozaev, line 76: Intentional code change during rewriting to typescript, looks safe for me.

packages/ui/src/ui/store/actions/navigation/modals/delete-object.tsx
```diff
@@ -399,15 +414,23 @@ function moveObjectsIntoTrash(multipleInfo, transaction, login) {
...
                 theme: 'success',
                 name: 'delete objects',
-                timeout: 10000,
-                title: 'Objects deleted',
-                content: 'Objects have been moved to the trash',
+                autoHiding: 10000,
+                title: 'Objects moved',
+                content: (
+                    <div>
+                        Objects have been moved to the trash, they could be found at
```
ma-efremoff, line 421: `they could be found at` -> `the can/might be found at`

ma-efremoff (approved):

PR #979, opened 2025-02-06
test(Navigation/Flow): add test for flow tab

KostyaAvtushko:
https://nda.ya.ru/t/ZyZW0WdX7BmEpG


ma-efremoff (approved):

ma-efremoff:
Next time it is better to create new nodes in `${E2E_DIR}/tmp`, then you will need to update less screenshots.

Issue #953, opened 2025-01-22
ManageTokens does not work for http


ma-efremoff:
At the moment our code uses [`crypto.suble`](https://github.com/ytsaurus/ytsaurus-ui/blob/df5efe5b0a9d74af8e52d935a17480794ad70437/packages/ui/src/ui/utils/sha256.ts#L2), by design it be might used only for secure contexts, see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle:
```
Secure context: This feature is available only in secure contexts (HTTPS),
in some or all [supporting browsers].
```

So we have to display corresponding notice for our users. Additionally we may provide `Password sha256` field instead of `Password` one when `crypto.subtle` is unavailable.

TODO:
- [ ]  Add explanation notice
- [ ] Provice `Password sha256` field instead of `Password`

ma-efremoff:
@vitshev

ma-efremoff:
Additionally we may provide `Password sha256` field when `crypto.subtle` is unavailable.

vrozaev:
PR with fix: https://github.com/ytsaurus/ytsaurus-ui/pull/980

PR #980, opened 2025-02-06
fix(ManageTokens): ManageTokens does not work for http [#953]

vitshev:


packages/ui/src/ui/containers/ManageTokens/ManageTokensPasswordModal/ManageTokensPasswordModal.tsx
```diff
@@ -41,26 +42,29 @@ const PasswordModal = (props: PasswordModalProps) => {
...
+                            username,
+                            ytAuthCluster,
+                            isSha256Mode(),
+                        );
+
+                        return strategy(sha256_password || password)
+                            .then((hashedPassword) => {
+                                props.handleConfirm(hashedPassword);
+                            })
+                            .catch((error: any) => setError(error?.response?.data || error));
```
ma-efremoff, line 57: It should also work if we call `setError(errro)`. Is there a reason to use error.response.data here?

packages/ui/src/ui/containers/ManageTokens/ManageTokensPasswordModal/ManageTokensPasswordModal.tsx
```diff
@@ -41,26 +42,29 @@ const PasswordModal = (props: PasswordModalProps) => {
...
                     fields={[
                         {
                             name: 'error-block',
                             type: 'block',
                             extras: {
-                                children: (
+                                children: isSha256Mode() ? (
                                     <Alert message="To access tokens management, you need enter your password" />
+                                ) : (
+                                    <Alert message="For security purposes, we only accept SHA-256 hashed passwords. Please hash your password before entering it." />
```
ma-efremoff, line 67: It is not about `security purposes`, it is about availability of `crypto.subtle`. It is better to mention it here:
```
crypto.subtle is not available from your browser, so SHA-256 hash cannot be generated. Please generate the hash of your password by self.
```
(`crypto.suptle` should be a link to `https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle`)

packages/ui/src/ui/containers/ManageTokens/ManageTokensPasswordModal/ManageTokensPasswordModal.tsx
```diff
@@ -41,26 +42,29 @@ const PasswordModal = (props: PasswordModalProps) => {
...
-                        });
+                        const {password, sha256_password} = data.getState().values;
+
+                        const strategy = createPasswordStrategy(
+                            username,
+                            ytAuthCluster,
+                            isSha256Mode(),
+                        );
+
+                        return strategy(sha256_password || password)
```
ma-efremoff, line 53:
```ts
const hashPromise = isCryptoSubleAvailable() ? sha256(password) : Promise.resolve(sha256_password);
return hashPromise.then(...);
```

packages/ui/src/ui/utils/sha256.ts
```diff
@@ -3,3 +3,7 @@ export async function sha256(str: string) {
 
     return [...new Uint8Array(buf)].map((x) => ('00' + x.toString(16)).slice(-2)).join('');
 }
+
+export const isSha256Mode = () => {
```
ma-efremoff, line 7: isCryptoSubleAvailable

packages/ui/src/ui/containers/ManageTokens/ManageTokensPasswordModal/password-strategies.ts
```diff
@@ -0,0 +1,32 @@
+import axios from 'axios';
```
ma-efremoff, line 1: It also may work without this file

ma-efremoff: From https://github.com/ytsaurus/ytsaurus-ui/pull/980#discussion_r1944671784
```
I would prefer to combine sha256 and calling login in a single method
```
It is a good reason to leave the file

ma-efremoff (approved):

Issue #768, opened 2024-09-15
UI column type search doesn't work with enter


Arustinal:
You can't use enter for data type selector in UI.
There's 2 problems - no autofocus on first search result and unexpected reaction for enter button. It's look like enter is applied to folders and tables, not create table menu.

How to reproduce:

1. Start create table from UI
2. Go to column selector
3. Try to select data type by typing into field
4. Hit enter

Kontakter:
@ma-efremoff – could you please take a look at this issue

vrozaev:
SIG UI Weekly discussion:

Yes, we do need fix that and check that we can use dialogs via keyboard, focus should not leave dialog.

And at least enter and escape should work.

vrozaev:
The main issue here is page hotkeys which still active when we are opening any dialog. We need to disable them when dialog became visible, that could be done with [scope change](https://github.com/jaywcjlove/hotkeys-js?tab=readme-ov-file#setscope).

vrozaev:
Fixed in https://github.com/ytsaurus/ytsaurus-ui/pull/992

PR #992, opened 2025-02-12
fix: prevent page hotkeys firing with opened YTDialog [#768]

vrozaev:
Issue: https://github.com/ytsaurus/ytsaurus-ui/issues/768

The rootcase of the issue is a way of hotkeys declaration via hotkeys-js.

Each hotkey is declared in some scope. Now that scope stays the same when any modal windows opens. That means that hotkeys from the page could be triggered with opened modal window, which might lead to unforeseen results.

The obvious solution here is switching hotkeys scope once any modal window become visible. That prevent page hotkeys from firing. 

So we need to modify all code which shows any king of modal windows and add there a scope switch.

1. packages/ui/src/ui/components/Dialog/Dialog.tsx
2. packages/ui/src/ui/components/Modal/Modal.tsx
3. packages/ui/src/ui/components/Modal/SimpleModal.tsx

Also, there are a lot of direct usage of `Dialog` and `Modal` components from `@gravity-ui/uikit`. I think we need to introduce wrappers for them and switch scope there. Let's name them `YTDialog` and `YTModal`.

### DIFF
--- a/packages/ui/.eslintrc
+++ b/packages/ui/.eslintrc
@@ -15,6 +15,20 @@
             "never"
         ],
         "lodash/import-scope": "error",
+        "no-restricted-imports": ["error", {
+            "paths": [
+                {
+                    "name": "@gravity-ui/uikit",
+                    "importNames": ["Dialog"],
+                    "message": "Please use src/components/YTDialog instead."
+                },
+                {
+                    "name": "@gravity-ui/uikit",
+                    "importNames": ["Modal"],
+                    "message": "Please use src/components/Modal or src/components/SimpleModal instead."
+                }
+            ]
+        }]
     },
     "env": {
         "jest": true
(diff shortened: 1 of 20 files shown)


packages/ui/.eslintrc
```diff
@@ -15,6 +15,20 @@
             "never"
         ],
         "lodash/import-scope": "error",
+        "no-restricted-imports": ["error", {
+            "paths": [
+                {
+                    "name": "@gravity-ui/uikit",
+                    "importNames": ["Dialog"],
```
ma-efremoff, line 22:
```ts
"name": "@gravity-ui/dialog-fields",
"importNames": ["DFDialog"],
```
And `// eslint-disable-next-line no-restricted-imports` should be also present for the rule

ma-efremoff (approved):

PR #1014, opened 2025-02-24
fix(Queries): now change of query ACO does not reset chart state and vice versa

vrozaev:


packages/ui/src/ui/pages/query-tracker/module/queryChart/actions.ts
```diff
@@ -23,24 +23,40 @@ import {
...
+const saveChartConfig = (
+    dispatch: Dispatch,
+    getState: () => RootState,
+    payload: SaveQueryChartConfigPayload,
+) => {
     dispatch(setSaved(false));
+    const state = getState();
+    const query = getQueryItem(state);
+    const isMultipleAco = selectIsMultipleAco(state);
+    const aco = {
```
SimbiozizV, line 46: use `getCurrentQueryACO` selector

packages/ui/src/ui/pages/query-tracker/module/api.ts
```diff
@@ -618,6 +622,8 @@ export function updateACOQuery({
         const state = getState();
         const isMultipleAco = selectIsMultipleAco(state);
         const {stage} = getQueryTrackerRequestOptions(state);
+        const query = selectQuery(state);
```
SimbiozizV, line 625: It is better to write new selectors for annotations. Similar to `getQueryId`

SimbiozizV (approved):

PR #1021, opened 2025-02-27
feat(Navigation/Queue): add features for queues exports [YTFRONT-4482]

KostyaAvtushko:
https://nda.ya.ru/t/wzTuoaTk7CdGtN

### DIFF
--- a/packages/ui/src/ui/hooks/use-updater.ts
+++ b/packages/ui/src/ui/hooks/use-updater.ts
@@ -3,7 +3,7 @@ import {useSelector} from 'react-redux';
 import isEqual_ from 'lodash/isEqual';
 
 import {Updater} from '../utils/hammer/updater';
-import {getUseAutoRefresh} from '../store/selectors/settings';
+import {getUseAutoRefresh} from '../store/selectors/settings/settings-ts';
 
 export const DEFAULT_UPDATER_TIMEOUT = 30 * 1000;
 
--- a/packages/ui/src/ui/store/api/yt/index.ts
+++ b/packages/ui/src/ui/store/api/yt/index.ts
@@ -3,7 +3,7 @@ import {BaseQueryFn, TypedUseMutationResult} from '@reduxjs/toolkit/dist/query/r
 
 import {rootApi} from '..';
 import {BatchApiArgs, BatchApiResults, executeBatchV3} from './endpoints/executeBatch';
-import {getUseAutoRefresh} from '../../../store/selectors/settings';
+import {getUseAutoRefresh} from '../../../store/selectors/settings/settings-ts';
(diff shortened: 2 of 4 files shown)


packages/ui/src/ui/pages/navigation/tabs/Queue/views/Exports/ExportsEditDialog/ExportsEditDialog.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    return (
+        <YTDFDialog<ExportsFormValues>
+            onAdd={onAdd}
+            visible={visible}
+            onClose={onClose}
+            size="l"
+            headerProps={{title: type === 'create' ? 'Create export' : 'Edit export'}}
+            initialValues={initialValues}
+            fields={fields}
```
ma-efremoff, line 59: There is no reason to keep fields in a separate file, the less hops we have the easier it is to read.
It is a good approach when reading of your code is linear without any hops and back-scrollings.

packages/ui/src/ui/pages/navigation/tabs/Queue/views/Exports/ExportsEditDialog/ExportsEditDialog.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    useEffect(() => {
+        const rawValues = configs ? configs[0].output : {};
+        setInitialValues(prepareInitialValues(rawValues, prevConfig?.export_name));
+    }, [prevConfig?.export_name, configs]);
+
+    const onAdd = async (form: FormApi<ExportsFormValues, Partial<ExportsFormValues>>) => {
+        const {values} = form.getState();
+        const preparedValues = prepareUpdateValues(values);
+        await update({
```
ma-efremoff, line 44: Is the promise resolved/rejected with result of corresponding request?

packages/ui/src/ui/store/api/navigation/tabs/queue/queue.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    if (type === 'edit' && newConfig?.export_name) {
+        newConfigs[newConfig?.export_name] = newConfig;
+    }
+
+    try {
+        const transactionId = await yt.v3.startTransaction({});
+        try {
+            await wrapApiPromiseByToaster(
```
ma-efremoff, line 98: We prefer to avoid using toaster for when it is possible. The mutation is used from dialog only, isn't it? If so we can display the error as part of the dialog.

ma-efremoff (approved):

ma-efremoff (approved):

PR #1056, opened 2025-03-21
minor fixes

ma-efremoff:
https://nda.ya.ru/t/tEdoELZT7CwXsc


packages/ui/src/ui/utils/unipika.ts
```diff
@@ -1,6 +1,14 @@
 import {UnipikaSettings} from '../components/Yson/StructuredYson/StructuredYsonTypes';
 import unipika from '../common/thor/unipika';
 
+export const YSON_AS_TEXT = (): UnipikaSettings => ({
```
KostyaAvtushko, line 4: why should it be a function? can it be just an object?

ma-efremoff: `unipika.formatFromYSON` and `unipika.formatRaw` injects defaults values to `settings` parameter, so it is required to avoid unnecessary changes of such contant.

KostyaAvtushko (approved):

PR #1057, opened 2025-03-24
feat(Queries/Graph): increased visibility of the number of tasks [YTFRONT-4812]

SimbiozizV:
https://nda.ya.ru/t/CdY8SIoo7D2g3G


ma-efremoff:
Don't you think it is better to use:
`feat(Queries/Graph): ...`  / `feat(Query/Graph): ...`
?

ma-efremoff (approved):

PR #1064, opened 2025-03-27
feat(Navigation): change title from tables [YTFRONT-4822]

SimbiozizV:
https://nda.ya.ru/t/c00lNx3n7D7DzJ


packages/ui/src/ui/pages/navigation/content/MapNode/MapNodesTable.js
```diff
@@ -74,8 +75,17 @@ export function renderMapNodesTableIcon(item) {
         icon = <Icon awesome={getIconNameForType(item.iconType, item.targetPathBroken)} />;
     }
 
+    let title = hammer.format['ReadableField'](item.type);
+    if (item.iconType === 'table') {
+        title = 'Static table';
+    }
+    if (item.iconType === 'table_dynamic') {
+        const {sorted} = ypath.getAttributes(item);
+        title = `${sorted ? 'Dynamic' : 'Queue'} table`;
```
KostyaAvtushko, line 84: no need for "table" after "Queue"

SimbiozizV: I don't agree. An unsorted table is still a table.
This is also one of the task conditions.

SimbiozizV: I asked the customer. Need with `table`

Arturaev: Yes, put the "table" as we agreed to use it in all places for queues.

KostyaAvtushko (approved):

PR #1124, opened 2025-05-12
fix(Navigation/Metadata): monaco vim mode [YTFRONT-4908]

SimbiozizV:


packages/ui/src/ui/components/EditTextWithPreview/EditTextWithPreview.tsx
```diff
@@ -54,6 +57,8 @@ export function EditTextWithPreview({
     disabled,
     initialSplitSize,
 }: EditTextWithPreviewProps) {
+    const vimMode = useSelector(getSettingsMonacoVimMode);
```
ma-efremoff, line 60: It is not allowed to use `redux`-related feature in `src/components` (like `useSelector`).
The component is used in several places of UI, but the option is available as `Settings/Queries/Vim mode`.

packages/ui/src/ui/components/EditTextWithPreview/EditTextWithPreview.scss
```diff
@@ -6,6 +6,7 @@
     min-height: 100px;
 
     &__left {
+        height: calc(100% - 40px);
```
ma-efremoff, line 9: 40px?

packages/ui/src/ui/pages/query-tracker/QueryEditor/QueryEditorView.scss
```diff
@@ -1,4 +1,6 @@
 .yt-qt-query-editor-view {
+    display: grid;
+    grid-template-rows: calc(100% - 40px) 40px;
```
ma-efremoff, line 3: 40px?

packages/ui/src/ui/components/EditTextWithPreview/EditTextWithPreview.tsx
ma-efremoff: As I understand this part is still actual:
`The component is used in several places of UI, but the option is available as Settings/Queries/Vim mode.`

ma-efremoff (approved):

PR #1125, opened 2025-05-12
feat(Analytics): analytics factory

vitshev:


packages/ui/src/ui/common/utils/metrics.ts
```diff
@@ -1,112 +1,54 @@
...
+    private async callImpl<K extends keyof AnalyticsService>(
+        method: K,
+        ...args: Parameters<AnalyticsService[K]>
+    ) {
+        await Promise.all(
+            this.services.map((service) =>
+                (service[method] as any)(...args).catch((e: Error) => {
+                    rumLogError(
+                        {
+                            message: `Failed to call ${method} in analytics service`,
```
ma-efremoff, line 41: It is better to add `${index}` at the end of the message

packages/ui/src/ui/UIFactory/default-ui-factory.tsx
```diff
@@ -319,4 +320,7 @@ export const defaultUIFactory: UIFactory = {
     renderMarkdown() {
         return undefined;
     },
+    getAnalyticsService(config) {
+        return [new YandexMetrika(config.metrikaCounterId)];
```
ma-efremoff, line 324: It is better to keep YandexMetrika working if `config.metrikaCounterId` is provided

ma-efremoff (approved):

PR #1140, opened 2025-05-20
fix: do not show cluster selectors in case when we have only one cluster

vrozaev:


Krock21 (approved):
nit: tests would be nice

ma-efremoff (approved):

PR #1144, opened 2025-05-20
feat(Navigation/Queue): add consumer creation [YTFRONT-4869]

KostyaAvtushko:


packages/ui/src/ui/pages/navigation/tabs/Consumer/modals/RegisterDialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    const clusters = Object.keys(
+        Object.fromEntries(
+            Object.entries(YT.clusters).filter(
+                ([_, value]) => value.environment === YT.clusters[cluster].environment,
+            ),
+        ),
+    );
+
+    const clusterControlOptions = clusters
```
SimbiozizV, line 39: Let's add useMemo

packages/ui/src/ui/pages/navigation/tabs/Consumer/modals/RegisterDialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+export function RegisterConsumerDialog() {
+    const dispatch = useDispatch();
+    const visible = useSelector(getRegisterDialogVisibility);
+    const cluster = useSelector(getCluster);
+
+    const onClose = () => dispatch(toggleRegisterDialog());
+
+    const [update, {isLoading, error}] = useRegisterMutation();
+
+    const clusters = Object.keys(
```
SimbiozizV, line 31: It seems to me that this way is more readable
```
const clusters = Object.entries(YT.clusters)
    .filter(([_, value]) => value.environment === YT.clusters[cluster].environment)
    .map(([key]) => key);
```

packages/ui/src/ui/pages/navigation/tabs/Queue/views/Consumers/RegisterConsumerDialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+export function RegisterConsumerDialog() {
+    const dispatch = useDispatch();
+    const visible = useSelector(getRegisterDialogVisibility);
+    const cluster = useSelector(getCluster);
+
+    const onClose = () => dispatch(toggleRegisterDialog());
+
+    const [update, {isLoading, error}] = useRegisterConsumerMutation();
+
+    const clusters = Object.keys(
```
SimbiozizV, line 35: I see this construction 2 times. Let's remove the duplication and move the code to helper

PR #1148, opened 2025-05-22
fix(Table): fix Preview YSON preview from `QT Kit` [YTFRONT-4952]

ma-efremoff:


packages/ui/src/ui/common/hammer/concat-validators.spec.js
```diff
@@ -0,0 +1,76 @@
...
+        result3 = null;
+        combinedValidator = concatValidators(validator1, validator2, validator3);
+    });
+
+    it('should return undefined when all validators return falsy values', () => {
+        result1 = null;
+        result2 = false;
+        result3 = '';
+
+        const value = 'test value';
```
SimbiozizV, line 23: repeated 3 times. Let's put it in a constant

PR #1187, opened 2025-06-09
feat(Navigation/Flow): add Graph tab [YTFRONT-3978]

ma-efremoff:
https://nda.ya.ru/t/S3FS3WYt7FFwEf

### DIFF
--- a/packages/ui/src/shared/yt-types.d.ts
+++ b/packages/ui/src/shared/yt-types.d.ts
@@ -587,7 +587,7 @@ export type FlowComputation = FlowNodeBase &
             memory_usage_current: number;
         };
         partitions_stats?: {
-            count: 1;
+            count: number;
             count_by_state?: Record<
                 'completed' | 'executing' | 'transient' | 'interrupted',
                 number | undefined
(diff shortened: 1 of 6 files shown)


packages/ui/src/ui/store/actions/flow/graph.ts
```diff
@@ -0,0 +1,33 @@
...
+import CancelHelper, {isCancelled} from '../../../utils/cancel-helper';
+import {flowGraphSlice} from '../../../store/reducers/flow/graph';
+
+type AsyncAction<R = void> = ThunkAction<R, RootState, unknown, any>;
+
+const cancelHelper = new CancelHelper();
+
+export function loadFlowGraph(pipeline_path: string): AsyncAction<Promise<void>> {
+    return (dispatch) => {
+        dispatch(flowGraphSlice.actions.onRequest({pipeline_path}));
```
KostyaAvtushko, line 14: Can we rewrite this with rtk-query api? I believe we can have a `flowExecute` and point, like `listQueries` 
https://github.com/ytsaurus/ytsaurus-ui/blob/main/packages/ui/src/ui/store/api/yt/endpoints/listQueries.ts

PR #1228, opened 2025-07-09
feat(Queries/List): infinite query list [YTFRONT-5060]

SimbiozizV:
https://nda.ya.ru/t/7rHi9Yof7GGCrQ


packages/ui/src/ui/pages/query-tracker/module/queries_list/queryListSlice.ts
```diff
@@ -0,0 +1,60 @@
...
+    listMode: QueriesListMode.History,
+};
+
+const queryListSlice = createSlice({
+    name: 'queryList',
+    initialState,
+    reducers: {
+        setLoading: (state, action: PayloadAction<boolean>) => {
+            state.isLoading = action.payload;
+        },
```
KostyaAvtushko, line 41: dont we want to try to rewrite it with rtk-query?

SimbiozizV: As far as I remember, according to the agreement, we switch to rtk-query at will.

evtaranov (approved):

PR #1244, opened 2025-07-28
fix(Accounts/DetailedUsage): move navigation link from path to button [YTFRONT-4896]

SimbiozizV:
https://nda.ya.ru/t/d_wLrkGv7Gwbs2
## Summary by Sourcery

Remove direct hyperlink wrapping the path in account detailed usage components and replace inline links with a dedicated external link button.

Bug Fixes:
- Remove the navigation link and tooltip icon from AccountActionsField so the path is no longer clickable.

Enhancements:
- Update AccountUsageDetails to render the path as plain text with a separate external link button and tooltip using UIKit Button and Icon instead of the custom Link component.

Chores:
- Clean up unused imports and remove the makeRoutedURL logic from AccountActionsField.


KostyaAvtushko:
can we reword pr title and commit, like: "move navigation link  from path to button"?

PR #1275, opened 2025-08-15
chore(PlaywrightComponents): fix tests

KostyaAvtushko:


SimbiozizV (approved):

packages/ui/src/ui/playwright-components/playwright.config.ts
```diff
@@ -40,7 +40,7 @@ const config: PlaywrightTestConfig = {
     /* Fail the build on CI if you accidentally left test.only in the source code. */
     forbidOnly: Boolean(process.env.CI),
     /* Retry on CI only */
-    retries: process.env.CI ? 2 : 0,
+    retries: 3,
```
ma-efremoff, line 43: It is ok to have retries for CI environment. But for local run it is better to avoid it.
So it is better to improve corresponding test, is it possible?

PR #1292, opened 2025-09-04
fix(Operation): utf8 error in description [YTFRONT-4982]

SimbiozizV:
https://nda.ya.ru/t/bFW9E0fb7JeSVU
       ## Summary by Sourcery

Disable UTF8 decoding in operation detail components to fix encoding errors when rendering descriptions and attributes, and add end-to-end tests with special characters to validate the fix.

Bug Fixes:
- Prevent UTF8 decoding errors in operation description and attributes by disabling decodeUTF8

Enhancements:
- Add getYsonSettingsWithoutDecode selector to clone YSON settings with decodeUTF8 disabled

Tests:
- Update init-cluster-e2e.sh to create operations containing special characters and add an e2e spec for operations


ma-efremoff:
Is there a reaso to split the PR for 3 commits?

PR #1311, opened 2025-09-19
fix(Operations/JobsMonitoring): full list of descriptors [YTFRONT-5192]

SimbiozizV:


packages/ui/src/ui/store/reducers/operations/jobs/jobs-monitor.ts
```diff
@@ -4,6 +4,7 @@ import {mergeStateOnClusterChange} from '../../../../store/reducers/utils';
 import {ActionD, YTError} from '../../../../types';
 
 export interface JobItem {
+    id: string;
```
KostyaAvtushko, line 7: what the motivation of the adding id attribute here?

PR #1342, opened 2025-10-23
fix(navigation/tablet-errors): fix showing errors for replicated tables

ainthero:
## Summary by Sourcery

Fix showing errors for replicated tables by updating the API batch request to retrieve the replicas map, summing per-replica error counts, and combining them with the table-level error count.

Bug Fixes:
- Fix tablet errors count for replicated tables by summing individual replica error_count fields and including table-level count.

Enhancements:
- Update API batch request to fetch full replicas map and handle missing or invalid replicas data.


ainthero:
> Can you please provide more details? Why do we need the fix?

In the current version, the "Errors" tab doesn’t appear for replicated tables when there are errors on replicas or tablets, unlike dyntables. You have to go to the "Tablets" tab, find the tablet with an error, open it, and only then you can see the error details. This fix resolves that issue.

PR #1503, opened 2026-04-08
feat(Queries/Navigation): allow to edit navigation path

ValeraS:
https://nda.ya.ru/t/GNpx50OO7Zbk3H
  ## Summary by Sourcery

Allow editing and navigation of the query tracker navigation path directly from the breadcrumbs header.

New Features:
- Enable inline editing of the navigation path via editable breadcrumbs and a path editor in the query tracker navigation header.

Enhancements:
- Update breadcrumbs items to generate proper routed URLs per segment and handle modified click behavior for opening paths.
- Normalize user-entered navigation paths to a canonical format before loading them.


packages/ui/src/ui/pages/query-tracker/Navigation/NavigationHeader/NavigationBreadcrumbs.tsx
```diff
@@ -33,41 +45,109 @@ export const NavigationBreadcrumbs: FC = () => {
...
         });
     }, [cluster, path, handleBreadcrumbsClick]);
 
     return (
         <>
-            <Button size="s" view="flat" onClick={handleClusterChangeClick}>
+            <Button view="flat" onClick={handleClusterChangeClick}>
                 <Icon data={FolderTreeIcon} size={16} />
             </Button>
-            <Breadcrumbs showRoot>{items}</Breadcrumbs>
```
xexys, line 70: `items.length > 0 &&`

packages/ui/src/ui/pages/query-tracker/Navigation/helpers/normalizePath.ts
```diff
@@ -0,0 +1,9 @@
+export function normalizePath(path: string) {
```
xexys, line 1: Rename `Navigation/helpers` to `Navigation/utils` ?

packages/ui/src/ui/pages/query-tracker/Navigation/helpers/normalizePath.ts
```diff
@@ -0,0 +1,9 @@
...
+    const trimmedPath = path.trim();
+
+    if (trimmedPath.match(/^\/*$/)) {
+        return '/';
+    }
+
+    return trimmedPath.replace(/^\/*/, '//').replace(/\/+$/, '');
+}
```
xexys, line 9: Can we add tests or a JSDoc @example so it’s clearer what this code is supposed to do?

packages/ui/src/ui/pages/query-tracker/Navigation/NavigationHeader/NavigationBreadcrumbs.scss
```diff
@@ -0,0 +1,10 @@
+.navigation-header-breadcrumbs {
+    &[class] {
```
xexys, line 3: Is &[class] here only to bump specificity and override another margin-right rule? If yes, can we add a short comment explaining that?

packages/ui/src/ui/pages/query-tracker/Navigation/NavigationHeader/NavigationBreadcrumbs.tsx
```diff
@@ -33,41 +45,109 @@ export const NavigationBreadcrumbs: FC = () => {
         [dispatch],
     );
 
+    const handleApplyPath = useCallback(
+        (newPath: string) => {
+            if (!clusterConfig) {
+                return;
+            }
+
+            const nextPath = normalizePath(newPath);
```
xexys: I suggest renaming:

- `nextPath` → `normalizedNewPath`

packages/ui/src/ui/pages/query-tracker/Navigation/NavigationHeader/NavigationBreadcrumbs.tsx
```diff
@@ -33,41 +45,109 @@ export const NavigationBreadcrumbs: FC = () => {
...
+                    onClick={(e) => {
+                        if (!e.ctrlKey && !e.metaKey && !e.shiftKey) {
+                            e.preventDefault();
+                            if (!isCurrent) {
+                                handleBreadcrumbsClick(item.path);
+                            }
+                        }
+                    }}
+                >
+                    {item.text}
```
xexys, line 115: Can we use new `RoutedLink` container instead?
https://github.com/ytsaurus/ytsaurus-ui/blob/main/packages/ui/src/ui/containers/RoutedLink/RoutedLink.tsx

xexys (changes requested):

packages/ui/src/ui/pages/query-tracker/Navigation/NavigationHeader/NavigationBreadcrumbs.tsx
ValeraS: `Breadcrumbs.Item` is already a link. The link here is only used to open a URL in a new window.

packages/ui/src/ui/pages/query-tracker/Navigation/NavigationHeader/NavigationBreadcrumbs.tsx
ValeraS: Why?

packages/ui/src/ui/pages/query-tracker/Navigation/helpers/normalizePath.ts
ValeraS: The folder `helpers` already exists and contains other files.

packages/ui/src/ui/pages/query-tracker/Navigation/NavigationHeader/NavigationBreadcrumbs.tsx
xexys: Simply a more compact recording that does the same thing :)

ValeraS: if it is only that then I prefer the current notation. :)

SimbiozizV:
Please update the branch. I want to build the farm and finally test that everything is in order.

SimbiozizV:
There are several problems in the implementation right now:

- Due to the narrow drop-down window, the path doesn't fit. This makes it difficult to understand where you want to go.
Снимок экрана 2026-04-16 в 15 50 13
Снимок экрана 2026-04-16 в 15 50 45

- I select something in the path list, but the transition doesn't happen. If I click anywhere outside the input field, the value resets to its original value. This is not obvious
- Autocomplete suggests the current path
Снимок экрана 2026-04-16 в 15 55 58

- I select something from the autosuggest to go there (for example, with the keyboard), and I press Enter. The transition doesn't happen. Only the autosuggest is hidden. I need to press Enter again to go to the selected path. However, the editing area doesn't close. This is also not obvious.
- Currently, path input is only available when a cluster is selected. Should path input be allowed along with the cluster?

ValeraS:
- The user can expand the left panel.

- The suggest behaviour is not implemented in this PR. This PR only reuses the existing suggest component, the same one already used on the navigation page. Because of that, I think, improving the UX of the suggest interaction itself is outside the scope of this PR.

- on the navigation page there are also two separate controls — one for selecting the cluster, and another for selecting the path inside that cluster. In most cases, the user already has the path inside the selected cluster in the clipboard and simply wants to paste it into the suggest field. Combining cluster and path in a single suggest input would make that workflow less convenient.

SimbiozizV:
> The user can expand the left panel.

Why not make the autosuggest field wider than the voode field?

Found an interesting bug. 

- Open the input field and start typing something like //home/
Снимок экрана 2026-04-16 в 16 59 27

- The autosuggest doesn't fit in width. I adjust the width of the navigation bar
Снимок экрана 2026-04-16 в 16 59 36

- Return to the input field, remove the / at the end of the path, and write it again to trigger the autosuggest
Снимок экрана 2026-04-16 в 17 04 59

- There is no autosuggest. It won't appear until I reopen the input field

SimbiozizV (approved):

xexys (approved):

PR #1513, opened 2026-04-14
fix(Operation/Details): show additional state only for running [YTFRONT-5709]

ma-efremoff:
https://nda.ya.ru/t/YcyEFfVg7ZW9Yf
 ## Summary by Sourcery

Bug Fixes:
- Ensure additional waiting status indicators are shown only when an operation is in the running state, preventing incorrect status display for other states.

### DIFF
--- a/packages/ui/src/ui/pages/operations/OperationDetail/OperationDetail.tsx
+++ b/packages/ui/src/ui/pages/operations/OperationDetail/OperationDetail.tsx
@@ -106,6 +106,10 @@ function getSpecialWaitingStatuses(
     type: string | undefined,
     isGpuOperation: boolean | undefined,
 ): {isWaitingForJobs?: boolean; isWaitingForResources?: boolean} {
+    if (state !== 'running') {
+        return {};
+    }
+
     const fairShareRatio = runtime?.[0]?.progress?.fair_share_ratio as number | undefined;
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/pages/operations/OperationDetail/OperationDetail.tsx
```diff
@@ -206,14 +206,10 @@ class OperationDetail extends React.Component<ReduxProps & RouteProps> {
...
-            state,
-            suspended,
-            runtime,
-            type,
-            isGpuOperation,
-        );
+        const {isWaitingForJobs, isWaitingForResources} =
+            state === 'running'
+                ? getSpecialWaitingStatuses(pools, state, suspended, runtime, type, isGpuOperation)
+                : {};
```
xexys: I suggest moving the condition `state === 'running'` to `getSpecialWaitingStatuses`, because the result of the function depends on the `state` one way or another, and it’s somehow strange to process a separate value outside the function.

xexys (changes requested):

packages/ui/src/ui/pages/operations/OperationDetail/OperationDetail.tsx
ma-efremoff: fixed

xexys (approved):

PR #1529, opened 2026-04-20
fix(Navigation/ReplicatedTable): add sorting in replicated tables and saving state in settings [YTFRONT-5703]

Lugertr:
https://nda.ya.ru/t/PAkpzYDL7aGpmB
       ## Summary by Sourcery

Add sortable columns and persistent sort state to the navigation replicated tables view.

New Features:
- Enable column-based sorting for replicated table entries, including separate sort options for replica ID and path in the name column.
- Persist replicated table sort preferences in global navigation settings and restore them when reopening or changing paths.

Enhancements:
- Integrate replicated table sorting with the shared tables sort-order state and assign a dedicated table ID.
- Adjust replicated table column widths to accommodate newly sortable content columns.


packages/ui/src/ui/pages/navigation/content/ReplicatedTable/ReplicatedTable.js
```diff
@@ -238,33 +257,131 @@ class ReplicatedTable extends Component {
...
+        const savedField = savedSortState.field || '';
+        const savedAsc = Boolean(savedSortState.asc);
+        const currentField = sortState?.field || '';
+        const currentAsc = Boolean(sortState?.asc);
+
+        if (currentField === savedField && currentAsc === savedAsc) {
+            return;
+        }
+
+        changeColumnSortOrder({
```
ma-efremoff: It is a bad idea to call such initializers. It is better to use corresponding selector.

packages/ui/src/ui/constants/navigation/content/replicated-table.ts
```diff
@@ -2,3 +2,5 @@ export const LOAD_REPLICAS_REQUEST = 'LOAD_REPLICAS_REQUEST';
 export const LOAD_REPLICAS_SUCCESS = 'LOAD_REPLICAS_SUCCESS';
 export const LOAD_REPLICAS_FAILURE = 'LOAD_REPLICAS_FAILURE';
 export const LOAD_REPLICAS_CANCELLED = 'LOAD_REPLICAS_CANCELLED';
+
+export const NAVIGATION_REPLICATED_TABLE_ID = 'navigation/replicated-table' as const;
```
ma-efremoff, line 6: Do we really need the constant?

packages/ui/src/ui/store/reducers/tables.ts
```diff
@@ -60,6 +62,7 @@ export const initialState: TablesSortOrderState = {
     [COMPONENTS_NODES_NODE_TABLE_ID]: {field: 'cell_id', asc: true},
 
     [COMPONENTS_PROXIES_TABLE_ID]: {field: 'host', asc: true},
+    [NAVIGATION_REPLICATED_TABLE_ID]: {field: 'cluster_name', asc: true},
```
ma-efremoff, line 65: Do we really need the field? Is it possible to pass sortState through props.

Lugertr: Yes, this field is necessary - in this component, the table is drawn using an ElementsTable,
this ElementsTable uses the toggleColumnSortOrder method for sorting, in it the sortState is taken from the state:

image

packages/ui/src/shared/constants/settings-types.ts
```diff
@@ -77,6 +77,8 @@ interface NavigationSettings {
     'global::navigation::queuePartitionsVisibility': Array<string>;
     'global::navigation::queueConsumersVisibility': Array<string>;
     'global::navigation::consumerPartitionsVisibility': Array<string>;
+
+    'global::navigation::replicatedTableSortState': {field: string; asc: boolean};
```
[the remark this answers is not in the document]
SimbiozizV: I agree with this comment. The user may not have this key set in their user settings. Let's consider the possibility of undefined in the type

SimbiozizV (dismissed):

SimbiozizV (approved):

PR #1531, opened 2026-04-21
fix(Navigation/tablets): correcting the sorting and display of fields…

Lugertr:
https://nda.ya.ru/t/ZqoMs_bv7a6And
    … [YT-5611]

## Summary by Sourcery

Fix sorting, filtering, and display of replication lag and mode columns in the Navigation tablets view.

Bug Fixes:
- Apply table sort and text filter to replication lag data before building histogram data.
- Ensure replication lag values are treated as numeric values and formatted as time durations.
- Normalize replication mode values for consistent sorting and display.

Enhancements:
- Align replication lag and replication mode column configurations with the generic table sorting/filtering utilities.

### DIFF
--- a/packages/ui/src/ui/constants/accounts/accounts.ts
+++ b/packages/ui/src/ui/constants/accounts/accounts.ts
@@ -118,7 +118,7 @@ export const ACCOUNT_RESOURCE_TYPES_DESCRIPTION = {
     [AccountResourceName.TABLET_STATIC_MEMORY]: {
         format: 'Bytes' as const,
         getInfo: (account: AccountSelector, recursive?: boolean): AccountResourceInfo => {
-            return account?.getTabletStaticMemoryInfo(recursive) || {};
+            return account?.selectTabletstaticMemoryInfo(recursive) || {};
         },
     },
     [AccountResourceName.MASTER_MEMORY]: {
--- a/packages/ui/src/ui/constants/accounts/accounts.ts
+++ b/packages/ui/src/ui/constants/accounts/accounts.ts
@@ -118,7 +118,7 @@ export const ACCOUNT_RESOURCE_TYPES_DESCRIPTION = {
     [AccountResourceName.TABLET_STATIC_MEMORY]: {
         format: 'Bytes' as const,
         getInfo: (account: AccountSelector, recursive?: boolean): AccountResourceInfo => {
-            return account?.selectTabletstaticMemoryInfo(recursive) || {};
+            return account?.selectTabletStaticMemoryInfo(recursive) || {};
         },
     },
     [AccountResourceName.MASTER_MEMORY]: {
--- a/packages/ui/src/ui/pages/accounts/selector.js
+++ b/packages/ui/src/ui/pages/accounts/selector.js
@@ -97,7 +97,7 @@ export default class Account {
         };
     }
 
-    getTabletStaticMemoryInfo(recursive = true) {
+    selectTabletstaticMemoryInfo(recursive = true) {
         const src = this.getResourceInfoSource(recursive);
         return {
             committed: src.committedTabletStaticMemory,
--- a/packages/ui/src/ui/pages/accounts/selector.js
+++ b/packages/ui/src/ui/pages/accounts/selector.js
@@ -97,7 +97,7 @@ export default class Account {
         };
     }
 
-    selectTabletstaticMemoryInfo(recursive = true) {
+    selectTabletStaticMemoryInfo(recursive = true) {
         const src = this.getResourceInfoSource(recursive);
         return {
             committed: src.committedTabletStaticMemory,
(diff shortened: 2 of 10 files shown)


packages/ui/src/ui/store/selectors/navigation/tabs/tablets.js
```diff
@@ -13,7 +13,34 @@ export const getTabletsMode = (state) => state.navigation.tabs.tablets.tabletsMo
...
+export const getTabletsSortState = (state) => state.tables[NAVIGATION_TABLETS_TABLE_ID];
+const getTabletsFilter = (state) => state.navigation.tabs.tablets.tabletsFilter;
+export const getActiveHistogram = (state) => state.navigation.tabs.tablets.histogramType;
+
+const getSortedReplicationLagTimes = createSelector(
+    [getRawReplicationLagTimes, getTabletsSortState],
+    (rawTablets, sortState) => hammer.utils.sort(rawTablets, sortState, tableItems),
+);
+
+const getFilteredReplicationLagTimes = createSelector(
```
SimbiozizV: Let's use the prefix select

packages/ui/src/ui/utils/navigation/tabs/tables.js
```diff
@@ -84,8 +84,9 @@ export const tableItems = {
         align: 'center',
         caption: 'Replication Mode',
         get(tablet) {
-            return ypath.getValue(tablet, '/replication_mode');
+            return ypath.getValue(tablet, '/replication_mode') || '';
```
SimbiozizV: In Tablets.renderReplicationMode, an empty value is still checked as typeof replicationMode !== 'undefined'. For '', this is true, which means that <Label … text="" /> will be used instead of hammer.format.NO_VALUE.

Is this the expected behavior?

packages/ui/src/ui/utils/navigation/tabs/tables.js
```diff
@@ -75,7 +75,7 @@ export const tableItems = {
         align: 'right',
         caption: 'Replication Lag',
         get(tablet) {
-            return ypath.getNumberDeprecated(tablet, '/replication_lag_time', undefined);
+            return Number(ypath.getValue(tablet, '/replication_lag_time'));
```
SimbiozizV: Has the chance of getting a NaN increased now? getNumberDeprecated handled such edge cases better

packages/ui/src/ui/store/selectors/navigation/tabs/tablets.js
Lugertr: oh, I forgot about naming, thanks

Lugertr: changed all the selectors in the file

packages/ui/src/ui/utils/navigation/tabs/tables.js
Lugertr: format then converts them correctly, but you're right - it's better to leave the previous option

packages/ui/src/ui/utils/navigation/tabs/tables.js
Lugertr: format then converts them correctly, but you're right - it's better to leave the previous option.

SimbiozizV (dismissed):

packages/ui/src/ui/pages/accounts/selector.js
```diff
@@ -97,7 +97,7 @@ export default class Account {
         };
     }
 
-    getTabletStaticMemoryInfo(recursive = true) {
+    selectTabletstaticMemoryInfo(recursive = true) {
```
ma-efremoff: It is incorrect to rename the method

packages/ui/src/ui/pages/navigation/tabs/Tablets/Tablets.js
```diff
@@ -24,10 +24,10 @@ import {getPath, getType} from '../../../../store/selectors/navigation';
 
 import {
     getActiveHistogram,
-    getNavigationTabletsLoadingStatus,
-    getTablets,
-    hasReplicationData,
     selectHistogram,
+    selectIsReplicationDataExist,
+    selectNavigationTabletsLoadingStatus,
+    selectTablets,
```
ma-efremoff, line 30: It is bad idea to mess renaming of selectors with fix in one commit. It is better to use separate PR for selectors-renaming.

packages/ui/src/ui/constants/accounts/accounts.ts
```diff
@@ -118,7 +118,7 @@ export const ACCOUNT_RESOURCE_TYPES_DESCRIPTION = {
     [AccountResourceName.TABLET_STATIC_MEMORY]: {
         format: 'Bytes' as const,
         getInfo: (account: AccountSelector, recursive?: boolean): AccountResourceInfo => {
-            return account?.getTabletStaticMemoryInfo(recursive) || {};
+            return account?.selectTabletstaticMemoryInfo(recursive) || {};
```
SimbiozizV: selectTabletstaticMemoryInfo -> selectTabletStaticMemoryInfo

SimbiozizV (approved):

PR #1554, opened 2026-04-30
feat(Navigation/Attributes): view opaque attributes [YTFRONT-2241]

xexys:
https://nda.ya.ru/t/XWpSwdMa7avkbP
                            ## Summary by Sourcery

Support on-demand loading and display of opaque navigation attributes in the Attributes tab.

New Features:
- Allow loading individual opaque attributes via per-row controls in the Attributes tab.
- Show toast-based error feedback when loading a single attribute fails.

Enhancements:
- Refine attributes Redux state to track overall and per-attribute request status, including opaque attribute keys.
- Adjust Attributes tab rendering to rely on the new request state and integrate extra row tools from Unipika.
- Update attribute selectors to expose opaque attribute keys and per-attribute loading metadata.

Build:
- Bump @gravity-ui/react-unipika to version 0.7.0.

Tests:
- Update navigation attributes E2E test selectors for the new filtered text CSS class.
- Add screenshot-based test artifacts for navigation opaque attributes.


packages/ui/src/ui/pages/navigation/tabs/Attributes/useRenderRowExtraTools.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import {useLoadableAttributesHandlers} from './useLoadableAttributesHandlers';
+
+export const useRenderRowExtraTools = (): RenderRowExtraTools => {
+    const {getIsPathLoadable, getLoadablePathState, onRequestLoadPath} =
+        useLoadableAttributesHandlers();
+
+    return useCallback(
+        ({path, value}) => {
+            if (!path || !value) {
```
ma-efremoff, line 15: Is it possible the value contains something like `false`, `0` or `""`?

packages/ui/src/ui/store/actions/navigation/tabs/attributes/attributes.ts
```diff
@@ -1,53 +1,92 @@
...
+            });
+
+            const data = await wrapApiPromiseByToaster(dataP, {
+                toasterName: 'load-attribute',
+                skipSuccessToast: true,
+                errorContent: i18n('alert_failed-to-load-attribute', {name}),
+            });
 
-            dispatch(attributesActions.onSuccess({attributes: output}));
+            dispatch(attributesActions.onAttributeSuccess({name, data}));
```
ma-efremoff, line 86: It is better to check current path before dispatch the action. In case when the path is changed the action should not be dispatched. Or path should be stored in the redux-slice and properly checked before save any data to the slice.

packages/ui/src/ui/pages/navigation/tabs/Attributes/useRenderRowExtraTools.tsx
xexys: No, here is always object!

packages/ui/src/ui/store/actions/navigation/tabs/attributes/attributes.ts
ma-efremoff: cancelHelper solves the issue.

ma-efremoff (approved):

PR #1559, opened 2026-05-04
fix(Components/Operations): redesign of operation's dialog of pools a…

Lugertr:
https://nda.ya.ru/t/I1ZlCfqK7bYUtK
                                             ## Summary by Sourcery

Redesign the operation pools and weights dialog to support resource limits and permission-aware editing.

New Features:
- Add a new React-based PoolsWeightsEditModal with tabbed editing for pools, weights, and resource limits, including a custom PoolWeightRow dialog control.
- Introduce per-tree resource limit handling for operations (CPU, GPU, memory, user slots) surfaced through selectors and the edit dialog.
- Add centralized PermissionsWarning component and permission checks for operation pool modifications.

Enhancements:
- Refine pool and weight update logic to only send changed fields and support resource_limits updates per pool tree.
- Improve dialog styling and layout for the operations pools/weights and pool editor dialogs, including better header/title handling.
- Add Redux state slice for operations page edit-weight modal, tracking loading, errors, visibility, editability, and permission check status.


packages/ui/src/ui/components/PermissionsWarning/PermissionsWarning.tsx
```diff
@@ -0,0 +1,31 @@
...
+export interface PermissionsWarningProps {
+    visible: boolean;
+    children?: React.ReactNode;
+    className?: string;
+}
+
+export function PermissionsWarning({visible, children, className}: PermissionsWarningProps) {
+    if (!visible) {
+        return null;
+    }
```
xexys: In most cases, this is a bad pattern.
The decision about which component to render should be made in the parent component.
The suggestion is to remove the `visible` prop.

packages/ui/src/ui/constants/operations/index.ts
```diff
@@ -8,8 +8,11 @@ export const OPERATIONS_STATUS = {
...
 
-export const SHOW_EDIT_WEIGHT_POOL_MODAL = PREFIX + 'SHOW_EDIT_WEIGHT_POOL_MODAL';
-export const HIDE_EDIT_WEIGHT_POOL_MODAL = PREFIX + 'HIDE_EDIT_WEIGHT_POOL_MODAL';
-export const SET_PULLS_AND_WEIGHTS = createActionTypes(PREFIX + 'SET_PULLS_AND_WEIGHTS');
+export const SHOW_EDIT_WEIGHT_POOL_MODAL = `${PREFIX}SHOW_EDIT_WEIGHT_POOL_MODAL` as const;
+export const HIDE_EDIT_WEIGHT_POOL_MODAL = `${PREFIX}HIDE_EDIT_WEIGHT_POOL_MODAL` as const;
+export const SET_POOLS_AND_WEIGHTS = createActionTypes(`${PREFIX}SET_POOLS_AND_WEIGHTS`);
+export const CHECK_OPERATION_PERMISSIONS = createActionTypes(
+    `${PREFIX}CHECK_OPERATION_PERMISSIONS`,
+);
```
xexys, line 20: Please add empty lines between exports

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.scss
```diff
@@ -1,26 +1,42 @@
...
+            font-size: 14px;
+            width: 200px;
+        }
     }
 
-    &__title {
-        justify-content: center;
-        display: flex;
-        align-items: center;
+    .g-dialog-header__caption {
```
xexys: This looks like a dirty fix because it breaks the component's encapsulation.

`g-dialog-header__caption` is not part of the component's public API.

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+const block = cn('operation-pools-weights');
+
+function makeError(error: unknown) {
+    return isEmpty_(error) ? null : (
+        <YTErrorBlock className={block('error')} error={error as Error} />
+    );
+}
+
+export default function PoolsWeightsEditModal() {
```
xexys: It is better not to use default export, especially for new components.

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+    const dispatch = useDispatch();
```
xexys, line 18: Please add empty line after

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+    const {
+        visible,
+        editable,
+        loading,
+        errorData,
+        operation,
+        hasWritePermission,
+        permissionLoading,
+        checkPermError,
+    } = useSelector((state: RootState) => state.operations.page.editWeightModal);
```
xexys: Maybe it would be better to move the selector code into a separate function?

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+        dispatch(hideEditPoolsWeightsModal());
+    }, [dispatch]);
+
+    const onAdd = useCallback(
+        async (form: FormApi<FormValues>) => {
+            const handler = buildSubmitHandler(operation, dispatch, setPoolsAndWeights);
+            await handler(form);
+        },
+        [dispatch, operation],
+    );
```
xexys: Suggestion: Move hooks to local modules

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+                    },
+                },
+                {
+                    name: 'memory',
+                    type: 'bytes' as const,
+                    caption: 'Memory',
+                    extras: {
+                        disabled: !editable,
+                    },
+                    validator(value?: number) {
```
xexys: Would be better `validator: (value?: number) {...}`, because there is no context dependency here

xexys: `validator(value: number | undefined)` ?

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+                        if (typeof value !== 'undefined' && isNaN(value)) {
```
xexys: `NaN` is always not equal `undefined`, so is first part of expression useless?

xexys: What's actually being checked here? Is the value an integer? Then that's the check that needs to be implemented.

packages/ui/src/ui/pages/operations/selectors.ts
```diff
@@ -80,13 +97,26 @@ export class OperationSelector implements Record<string, any> {
             const isEphemeral = ypath.getValue(treeData, '/running_in_ephemeral_pool') || false;
             const isLightweight = ypath.getValue(treeData, '/running_in_lightweight_pool') || false;
 
+            const rawLimits = schedulingInfo.resource_limits;
+            let resourceLimits: OperationPoolResourceLimits | undefined;
+            if (rawLimits && typeof rawLimits === 'object') {
```
xexys, line 103: It is advisable to add an empty line before id or before the `resourceLimits` variable

packages/ui/src/ui/components/PermissionsWarning/PermissionsWarning.tsx
```diff
@@ -0,0 +1,31 @@
...
+                <>
+                    You do not have sufficient permissions to modify pool settings. In order to be
+                    able to modify pool settings <span className={block('flag')}>inherit_acl</span>{' '}
+                    flag should be enabled for current pool and you should be responsible for the
+                    parent pool.
+                </>
+            )}
+        </div>
+    );
+}
```
xexys: This API looks strange. A better solution is to create a main component that accepts some text, and several specific components based on the main one.

```ts
const PermissionsWarning = ({children}) => <div>{children}</div>

const FooPermissionsWarning = () => <PermissionsWarning>Foo</PermissionsWarning>

const BarPermissionsWarning = () => <PermissionsWarning>Bar</PermissionsWarning>

```

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.scss
xexys: Same story about `df-dialog` and `df-tab-field-vertical` classNames

xexys (changes requested):

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/utils.tsx
```diff
@@ -0,0 +1,68 @@
...
+    user_slots: NumberFieldValue;
+};
+
+export type FormValues = Record<string, PoolData>;
+
+export function buildInitialValues(pools: OperationPool[]): FormValues {
+    const formValues: FormValues = {};
+
+    for (const item of pools) {
+        const rl = item.resourceLimits || {};
```
xexys: It is better not to use abbreviations if possible.
Can destructuring be used here?

xexys (changes requested):

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
Lugertr: this check need to avoid ts error:

image

packages/ui/src/shared/utils/toolkit/type/isPositiveOrZeroNumber.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Проверяет, что значение является неотрицательным числом.
```
SimbiozizV: don't forget that we leave comments in English

packages/ui/src/ui/components/PermissionsWarning/PermissionsWarning.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const PermissionsWarning = ({children, className}: PermissionsWarningProps) => {
+    return <div className={block(null, className)}>{children}</div>;
+};
+
+export const PoolPermissionsWarning = ({className}: {className?: string}) => {
+    return (
+        <PermissionsWarning className={className}>
+            You do not have sufficient permissions to modify pool settings. In order to be able to
```
SimbiozizV: Let's move this to i18n

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+        hasWritePermission,
+        permissionLoading,
+        checkPermError,
+    } = useSelector((state: RootState) => state.operations.page.editWeightModal);
+
+    const user = useSelector(selectCurrentUserName);
+
+    const showPermissionWarning = hasWritePermission === false;
+
+    const pools = useMemo(() => operation.pools ?? [], [operation.pools]);
```
SimbiozizV: This variable is only used in initialValue. Its calculation doesn't look heavy
 I suggest moving it to the useMemo variable initialValues

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+            await handler(form);
+        },
+        [dispatch, operation],
+    );
+
+    if (!visible || !pools.length) {
+        return null;
+    }
+
+    const titlePrefix = 'Edit Operations tools';
```
SimbiozizV: i18n

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+            <div>{titlePrefix}</div>
+            <Tooltip placement={'bottom'} content={headerTitleSuffix}>
+                <span className={block('title-suffix')}>{headerTitleSuffix}</span>
+            </Tooltip>
+        </div>
+    ) : (
+        titlePrefix
+    );
+
+    const poolTabs = pools.map((item) => {
```
SimbiozizV: This is not an error, but a suggestion. But I would remove this huge block from the component body

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/utils.tsx
```diff
@@ -0,0 +1,68 @@
...
+            gpu: {value: gpu},
+            memory,
+            user_slots: {value: user_slots},
+        };
+    }
+
+    return formValues;
+}
+
+export function buildSubmitHandler(
```
SimbiozizV: I don't like that dispatch is passed in the function arguments.
I suggest that you form the data here and do the dispatch at the point of use

packages/ui/src/ui/components/PermissionsWarning/PermissionsWarning.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import cn from 'bem-cn-lite';
+
+import './PermissionsWarning.scss';
+
+const block = cn('yt-permissions-warning');
+
+export interface PermissionsWarningProps {
+    children?: React.ReactNode;
```
xexys: is it really optional?

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+        checkPermError,
+    } = useSelector((state: RootState) => state.operations.page.editWeightModal);
+
+    const user = useSelector(selectCurrentUserName);
+
+    const showPermissionWarning = hasWritePermission === false;
+
+    const pools = useMemo(() => operation.pools ?? [], [operation.pools]);
+    const isTerminal = OPERATION_TERMINAL_STATES.has(
+        operation.state as Parameters<typeof OPERATION_TERMINAL_STATES.has>[0],
```
xexys: This type is more readable and expresses the meaning better.
`ValuesType<typeof OPERATION_TERMINAL_STATES>`

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+            headerProps={{title: dialogTitle}}
+            footerProps={editable ? undefined : {hidden: true}}
+            isApplyDisabled={(state) =>
+                !editable ||
+                state.pristine ||
+                state.hasValidationErrors ||
+                loading ||
+                isTerminal ||
+                permissionLoading
+            }
```
xexys, line 68: Is it really necessary to override the default function? 
Does it work incorrectly out of the box?

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/utils.tsx
xexys: I agree, it's better to replace the function with a hook that returns the handler.

packages/ui/src/ui/store/actions/operations/index.js
```diff
@@ -26,37 +30,106 @@ export function hideEditPoolsWeightsModal() {
...
+        }
+
+        dispatch({type: CHECK_OPERATION_PERMISSIONS.REQUEST});
+
+        const hashPrefix = hammer.utils.extractFirstByte(operationValue);
+        const operationPath = `//sys/operations/${hashPrefix}/${operationValue}`;
+
+        checkUserPermissionsUI(operationPath, user, ['manage'])
+            .then((results) => {
+                const hasWritePermission = results.every((r) => r.action === 'allow');
```
xexys: I suggest not using abbreviations whenever possible.

`results.every(({action}) => action === 'allow');`

packages/ui/src/ui/store/actions/operations/index.js
```diff
@@ -26,37 +30,106 @@ export function hideEditPoolsWeightsModal() {
...
+    };
+}
+
+function diffResourceLimits(oldLimits = {}, newLimits = {}) {
+    let changed = false;
+    const merged = {};
+
+    for (const key of OPERATION_POOL_RESOURCE_LIMIT_KEYS) {
+        const hasNewValue = Object.hasOwnProperty.call(newLimits, key);
+        const hasOldValue = Object.hasOwnProperty.call(oldLimits, key);
```
xexys: Is it really necessary to build logic based on the presence of a key in an object?

Such logic is usually extremely fragile; it's better to rely on values ​​whenever possible.

packages/ui/src/ui/store/actions/operations/index.js
```diff
@@ -26,37 +30,106 @@ export function hideEditPoolsWeightsModal() {
...
+            merged[key] = null;
+        }
+
+        if (oldVal !== newVal || hasOldValue !== hasNewValue) {
+            changed = true;
+        }
+    }
+
+    return {changed, merged};
+}
```
xexys: Are there any similar functions in lodash? Or can I achieve the same result by combining lodash functions?

packages/ui/src/ui/store/reducers/operations/page.ts
```diff
@@ -0,0 +1,181 @@
...
+    | HideAction
+    | SetRequestAction
+    | SetSuccessAction
+    | SetCancelledAction
+    | SetFailureAction
+    | CheckRequestAction
+    | CheckSuccessAction
+    | CheckFailureAction;
+
+export default function operationsPageReducer(
```
xexys: It is better not to use default export if possible.

xexys (changes requested):

packages/ui/src/ui/components/PermissionsWarning/PermissionsWarning.tsx
Lugertr: this parameter is not actually used, it's better to delete it.

packages/ui/src/ui/store/reducers/operations/page.js
ma-efremoff, line 1: The file should be moved not deleted

packages/ui/src/shared/utils/toolkit/type/isPositiveOrZeroNumber.ts
```diff
@@ -0,0 +1,19 @@
...
+ * isPositiveOrZeroNumber(42) // true
+ * isPositiveOrZeroNumber(42.5) // true
+ * isPositiveOrZeroNumber(Infinity) // true
+ * isPositiveOrZeroNumber(0) // true
+ * isPositiveOrZeroNumber(-1) // false
+ * isPositiveOrZeroNumber('42') // false
+ * isPositiveOrZeroNumber(NaN) // false
+ * ```
+ */
+export const isPositiveOrZeroNumber = (value: unknown): value is number => {
```
ma-efremoff: We already have `validateNumber` helper function

packages/ui/src/ui/pages/scheduling/Scheduling/PoolEditorDialog/PoolEditorDialog.scss
```diff
@@ -29,4 +29,4 @@
     &__transfer-quota-notice {
         padding-bottom: 20px;
     }
-}
+}
```
xexys: It looks like there is an extra line here?

packages/ui/src/ui/store/reducers/operations/page.ts
```diff
@@ -0,0 +1,181 @@
...
+                    editable,
+                    operation,
+                },
+            };
+        }
+
+        case HIDE_EDIT_WEIGHT_POOL_MODAL:
+            return {
+                ...state,
+                editWeightModal: {...initialState.editWeightModal},
```
xexys: it is not necessary to use the spread in this place

xexys (changes requested):

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/PoolsWeightsEditModal.tsx
Lugertr: make more readable - `as OperationStates`

packages/ui/src/shared/utils/toolkit/type/isPositiveOrZeroNumber.ts
ma-efremoff: we already have `validateNumber` helper

packages/ui/src/ui/constants/operations/index.ts
```diff
@@ -2,14 +2,21 @@ import createActionTypes, {createPrefix} from '../../constants/utils';
 import {Page} from '../../constants/index';
 
 const PREFIX = createPrefix(Page.OPERATIONS);
+
```
ma-efremoff, line 5: It looks like some changes in the file should be in previous commit

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/utils.tsx
```diff
@@ -0,0 +1,115 @@
...
+        if (!poolData.pool || !poolData.tree) {
+            continue;
+        }
+
+        poolsMap[tree] = poolData.pool;
+        weightsMap[tree] = poolData.weight?.value;
+
+        const resourceLimits: OperationPoolResourceLimits = {};
+
+        if (poolData.cpu?.value !== undefined) resourceLimits.cpu = poolData.cpu.value;
```
ma-efremoff: What will we have in `poolData.cpu.value` if user removes limit for CPU? ?

packages/ui/src/ui/pages/operations/PoolsWeightsEditModal/usePoolTabs.tsx
```diff
@@ -0,0 +1,111 @@
...
+import i18n from './i18n';
+
+type UsePoolTabsParams = {
+    pools: OperationPool[];
+    editable: boolean;
+};
+
+export function usePoolTabs({pools, editable}: UsePoolTabsParams) {
+    return useMemo(() => {
+        return pools.map((item) => {
```
ma-efremoff: There is no reason to extract properties to a separate file. Inplace props definition for YTDFDialog is good enough.

packages/ui/src/ui/pages/scheduling/Scheduling/PoolEditorDialog/PoolEditorDialog.tsx
```diff
@@ -470,8 +471,8 @@ export function PoolEditorDialog() {
                             type: 'bytes',
                             caption: 'Memory',
                             validator(value?: number) {
-                                if (typeof value !== 'undefined' && isNaN(value)) {
-                                    return i18n('alert_incorrect-value');
+                                if (value && !isPositiveOrZeroNumber(value)) {
```
ma-efremoff: Is everything right here?

packages/ui/src/ui/pages/scheduling/Scheduling/PoolEditorDialog/PoolEditorDialog.tsx
```diff
@@ -470,8 +471,8 @@ export function PoolEditorDialog() {
...
+                                    i18n('alert_incorrect-value');
```
ma-efremoff: ?

packages/ui/src/shared/utils/toolkit/type/isPositiveOrZeroNumber.ts
Lugertr: For the most part, this check is needed to validate that the number is valid and greater than zero,
but I agree, the record looks strange.

I used this function, not `validateNumber`, because `validateNumber` does not have i18n multilingualism.

Ok, I will add validateNumber

packages/ui/src/ui/store/reducers/operations/page.js
Lugertr: it turned out to be just a separate commit to import the file, otherwise git does not see it.

ma-efremoff (dismissed):

xexys (approved):

PR #1568, opened 2026-05-12
fix(ACL): pass `name` field to `updateAcl` [YTFRONT-5762]

ma-efremoff:
https://nda.ya.ru/t/c68Wry-e7cFYUU
  ## Summary by Sourcery

Update ACL handling to pass object name and integrate responsible user assignment via Redux actions for accounts and pools.

New Features:
- Add Redux thunk to create accounts and set responsible users through the centralized ACL update action.
- Support specifying responsible users and ABC service when creating scheduling pools via the create pool dialog.

Bug Fixes:
- Ensure ACL updates include the name field when calling the ACL API to fix missing-name issues.

Enhancements:
- Refactor account and pool creation flows to use shared ACL update actions instead of direct ACL API calls.
- Strongly type create pool and account creation forms with explicit ResponsibleType and ABC service structures.

### DIFF
--- a/packages/ui/src/ui/store/actions/scheduling/create-pool-dialog.tsx
+++ b/packages/ui/src/ui/store/actions/scheduling/create-pool-dialog.tsx
@@ -1,28 +1,28 @@
 import React from 'react';
 import {type ThunkAction} from 'redux-thunk';
+import {
+    CREATE_POOL_DIALOG_TREE_CREATE_FAILURE,
+    CREATE_POOL_DIALOG_TREE_ITEMS_SUCCESS,
+} from '../../../constants/scheduling';
 import {type RootState} from '../../reducers';
 import {
     type CreatePoolDialogAction,
     type CreatePoolDialogState,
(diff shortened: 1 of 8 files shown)


packages/ui/src/ui/store/actions/scheduling/create-pool-dialog.tsx
```diff
@@ -71,11 +76,9 @@ export function createPool(values: FIX_MY_TYPE): CreatePoolDialogThunkAction {
             },
         )
             .then(() => {
-                updateAcl(cluster, values.name, {
-                    idmKind: IdmObjectType.POOL,
-                    poolTree: values.tree,
-                    responsible: values.responsible,
-                });
+                if (UIFactory.getAclApi().isAllowed) {
+                    dispatch(updateAcl({values, path: values.name, idmKind: IdmObjectType.POOL}));
```
SimbiozizV: There is redundant data in value. Let's only pass what the api needs

SimbiozizV (approved):

PR #1590, opened 2026-05-19
fix(Scheduling): allow any pool name suggestion

Lugertr:
https://nda.ya.ru/t/BWd2EQRF7cib7i
                                ## Summary by Sourcery

Handle missing or inaccessible scheduling pools more gracefully and allow arbitrary pool names to be entered and preserved in the UI.

Bug Fixes:
- Prevent automatic fallback to the root pool when a requested pool cannot be resolved and instead show an appropriate pool-not-found error state.
- Ensure scheduling breadcrumbs and navigation correctly reflect and preserve an invalid or erroring pool without allowing navigation into it.
- Allow pool selection to accept and apply arbitrary text input rather than restricting to suggested pool names.

Enhancements:
- Display a dedicated pool-not-found error view with richer visuals, details, and copy-to-clipboard support when pool loading fails with known error codes.
- Simplify scheduling reducer state by removing unused poolLoaded, poolError, and isNewPools flags and centralizing error handling via poolErrorData and a new SCHEDULING_POOL_NOT_FOUND action.


packages/ui/src/ui/components/Suggest/Suggest.tsx
```diff
@@ -67,6 +67,7 @@ export interface SuggestProps {
     };
     items: Array<SuggestItem>;
     maxItems?: number | {totalAmount: number; groupPredicate: () => void};
+    allowArbitraryInput?: boolean;
```
ma-efremoff: Why do we need the property? Isn't it enough to have existing `onTextUpdate`/`onKeyDown`?

packages/ui/src/ui/pages/scheduling/Content/Content.tsx
```diff
@@ -114,35 +117,47 @@ function Content({match, location}: ContentProps) {
...
-                        return <Redirect to={makeSchedulingUrl({pool, poolTree: tree})} />;
-                    }}
-                />
-                <Route path={`${match.path}/:tab`} component={Placeholder} />
-                <Redirect
-                    from={match.url}
-                    to={{pathname: `${match.url}/${DEFAULT_TAB}`, search: location.search}}
-                />
-            </Switch>
+            {poolNotFound ? (
```
ma-efremoff: What will be displayed without the check?

packages/ui/src/ui/store/reducers/scheduling/scheduling.ts
```diff
@@ -92,6 +94,7 @@ const ephemeralState: SchedulingEphemeralState = {
     poolLoading: false,
     poolLoaded: false,
     poolError: false,
+    poolNotFound: false,
```
ma-efremoff: It is better to avoid such additional field in the state. All we need should be present in corresponding error-field.

packages/ui/src/ui/pages/scheduling/Content/PoolNotFoundError.tsx
```diff
@@ -0,0 +1,15 @@
...
+import {NoContent} from '@ytsaurus/components';
+import i18n from './i18n';
+
+export const PoolNotFoundError = ({poolName}: {poolName?: string}) => {
+    const warningText = i18n('pool_not_found', {poolName});
+    const hintText = i18n('pool_not_found_hint', {poolName});
+
+    return (
+        <Flex justifyContent="center" grow={1}>
+            <NoContent warning={warningText} padding="large" hint={hintText} />
```
ma-efremoff: We use `NoContent` when we expect some choice from user. Incorrect choice is an error that should be displayed properly. A wrong pool-choice should be displayed similar to a wrong path in navigation.

packages/ui/src/ui/components/Suggest/Suggest.tsx
Lugertr: The `allowArbitraryInput` property allows users to input text that doesn't match any of the provided suggestions

the standard behavior (onTextUpdate/onKeyDown) does not imply this - I do not want to break the logic, since this is a component that is reused in many other components.

packages/ui/src/ui/pages/scheduling/Content/Content.tsx
Lugertr: now i am change check to 'poolErrorData', which have inside any error of request,

without check will be just empty tabs with empty tables of scheduling

packages/ui/src/ui/store/reducers/scheduling/scheduling.ts
Lugertr: I have explored all the additional fields and deleted those that are not in use. (poolNotFound too)

packages/ui/src/ui/pages/scheduling/Content/PoolNotFoundError.tsx
Lugertr: made 404 mostly as in Navigation
(it only hasn't error title as in navigation, can add it if it needs)

image

packages/ui/src/ui/components/Suggest/Suggest.tsx
ma-efremoff: Suggest control by default allows any input from users

packages/ui/src/ui/pages/scheduling/Content/helpers/helpers.ts
```diff
@@ -0,0 +1,34 @@
+import {getYtErrorCode} from '../../../../utils/errors';
+import {type YTError} from '../../../../../@types/types';
+
+export type PoolErrorCode = 500 | 901;
+
+function collectErrorCodes(error: YTError, dst = new Set<number>()) {
```
ma-efremoff: Why do we need to duplicate the functions?

packages/ui/src/ui/pages/scheduling/Content/PoolNotFoundError.tsx
```diff
@@ -0,0 +1,77 @@
...
+    );
+}
+
+function UnexpectedPoolError(props: Props) {
+    const {details, message} = props;
+
+    return <YTErrorBlock className={block('unexpected-error')} error={details} message={message} />;
+}
+
+export function PoolNotFoundError(props: Props) {
```
ma-efremoff: Is it possible to avoid copy-paste of NavigationError component?

packages/ui/src/ui/components/Suggest/Suggest.tsx
Lugertr: by input, I meant submit (in the task, when you click enter, you were asked to allow the input of a value that is not in the suggestion)

packages/ui/src/ui/pages/scheduling/Content/PoolNotFoundError.tsx
Lugertr: move mostly all copy-paste to a separate component

packages/ui/src/ui/components/Suggest/Suggest.tsx
```diff
@@ -72,6 +72,7 @@ export interface SuggestProps {
     onBlur?: () => void;
     onFocus?: () => void;
     onKeyDown?: (e: React.KeyboardEvent<HTMLElement>) => void;
+    onEnterKeyDown?: (value: string, e: React.KeyboardEvent<HTMLElement>) => void;
```
xexys: Suggestion: Do not use abbreviations

packages/ui/src/ui/components/Suggest/Suggest.tsx
```diff
@@ -196,6 +197,7 @@ export default class Suggest extends Component<SuggestProps, State> {
             evt.preventDefault();
             this.selectPrevSuggestion();
         } else if (key === ENTER) {
+            this.props.onEnterKeyDown?.(this.state.text ?? '', evt);
```
xexys: Suggestion: Do not use abbreviations

packages/ui/src/ui/containers/PrettyError/helpers/error-helpers.ts
```diff
@@ -0,0 +1,78 @@
...
+    permissions?: Array<UnipikaValue>;
+    path?: string;
+    [key: string]: any;
+};
+
+export type ErrorsInfoMap = {
+    [key in ErrorCode]?: {
+        getTitle: (payload: ErrorTitlePayload) => string;
+    };
+};
```
xexys: You can use `Record` here - same result, but more readable

packages/ui/src/ui/containers/PrettyError/helpers/error-helpers.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    return dst;
+}
+
+export function determineErrorCode(error: YTError): ErrorCode | undefined {
+    const codes = collectErrorCodes(error);
+
+    if (codes.has(901)) return 901;
```
xexys: It is better not to use if statement without brackets

packages/ui/src/ui/containers/PrettyError/helpers/error-helpers.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (codes.has(500)) return 500;
+
+    return undefined;
+}
+
+export function findErrorByCode(error: YTError, code: ErrorCode): YTError | undefined {
+    const currentCode = getYtErrorCode(error);
+    if (currentCode === code) return error;
```
xexys: It is better not to use if statement without brackets

packages/ui/src/ui/containers/PrettyError/helpers/error-helpers.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    if (error?.inner_errors) {
+        for (const inner of error.inner_errors) {
+            const found = findErrorByCode(inner, code);
+            if (found) return found;
```
xexys: It is better not to use if statement without brackets

packages/ui/src/ui/containers/PrettyError/helpers/error-helpers.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const getTitle = errorsInfo[code]!.getTitle;
+
+    const targetError = findErrorByCode(error, code) || error;
+    const attributes = targetError.attributes;
+
+    return getTitle({
+        username: attributes ? ypath.getValue(attributes, '/user') : '',
+        permissions: attributes ? ypath.getValue(attributes, '/permission') : [],
+        ...contextPayload,
```
xexys: When the spread comes at the end, it usually means we want to define default values ​​that can be overwritten.

If this isn't the case in this case, it's better to create the spread first and then specify specific fields.

packages/ui/src/ui/containers/PrettyError/PrettyError.tsx
```diff
@@ -0,0 +1,83 @@
...
+    errorCode?: ErrorCode;
+    errorsInfo?: ErrorsInfoMap;
+    errorContext?: Partial<ErrorTitlePayload>;
+    renderActions?: (error: YTError) => React.ReactNode;
+    renderAdditionalContent?: () => React.ReactNode;
+    vertical?: boolean;
+    className?: string;
+};
+
+function PrettyErrorView(props: PrettyErrorProps & {code: ErrorCode; title: string}) {
```
xexys: It's better to move props into a separate data type.

packages/ui/src/ui/containers/PrettyError/PrettyError.tsx
```diff
@@ -0,0 +1,83 @@
...
+                            {i18n('action_copy-error-details')}
+                        </ErrorToClipboardButton>
+                    </Flex>
+                </Flex>
+            </Flex>
+        </Flex>
+    );
+}
+
+export function PrettyError(props: PrettyErrorProps) {
```
xexys, line 76: The component path is `packages/ui/src/ui/containers/PrettyError/PrettyError.tsx`

Is it really common component for renders any errors? 

Error code 901 is used here. What is this code? It doesn't look like a generic application-level code.
It looks like it's a specific error code for a specific error. This shouldn't be in a generic error rendering component.

packages/ui/src/ui/pages/flow/flow-components/FlowError/FlowError.tsx
```diff
@@ -9,7 +9,5 @@ export function FlowError({error}: {error?: YTError}) {
     const path = useSelector(selectFlowPipelinePath);
     const cluster = useSelector(selectCluster);
 
-    return !error ? null : (
-        <NavigationError details={error} path={path} cluster={cluster} message={error.message} />
-    );
+    return !error ? null : <NavigationError details={error} path={path} cluster={cluster} />;
```
xexys, line 12: It's best to avoid rendering components that return `null` if possible

Make the `error` condition mandatory, and move the component's render condition to the parent component where the call occurs.

packages/ui/src/ui/store/selectors/scheduling/scheduling.ts
```diff
@@ -58,6 +58,11 @@ export const getTrees = (state: RootState) => state.scheduling.scheduling.trees;
 
 export const getPool = (state: RootState) => state.scheduling.scheduling.pool;
 
+export const selectSchedulingError = (state: RootState) => {
+    const {error: hasError, errorData} = state.scheduling.scheduling;
```
xexys: error: hasError - It looks like a misnamed rename that only confuses

xexys (changes requested):

packages/ui/src/ui/components/Suggest/Suggest.tsx
```diff
@@ -186,16 +187,17 @@ export default class Suggest extends Component<SuggestProps, State> {
         this.setState({focused: false});
     }
 
-    onKeyDown(evt: React.KeyboardEvent<HTMLElement>) {
-        const key = evt.keyCode;
+    onKeyDown(event: React.KeyboardEvent<HTMLElement>) {
+        const key = event.keyCode;
```
xexys, line 191: Is renaming really necessary here? Why not just use the `keyCode` variable?

`const {keyCode} = event;`

packages/ui/src/ui/pages/scheduling/Scheduling/SchedulingError.tsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+type Props = {
+    error: YTError;
+    poolName?: string;
+    className?: string;
+};
+
+export function SchedulingError({error, poolName, className}: Props) {
+    const errorCode = determineErrorCode(error);
```
ma-efremoff: `determineErrorCode` may also return `901`, but ErrorsInfo contains only `500`.
Is it correct behavior?

packages/ui/src/ui/store/actions/scheduling/expanded-pools.tsx
```diff
@@ -135,30 +133,26 @@ export function loadExpandedPools(tree: string): ExpandedPoolsThunkAction {
...
-                         */
-                        // dispatch(loadExpandedOperationsAndPools(tree));
-                    } else {
-                        toaster.add({
-                            name: 'schedulingPoolFullPath',
-                            theme: 'danger',
-                            title: '',
-                        });
-                    }
+                    dispatch(poolNotFound(e));
```
ma-efremoff: Is it correct that we always dispatch `poolNotFound` here?
Is it possible to get another type of error?

Lugertr: renamed the method

ma-efremoff (approved):

xexys (changes requested):

xexys (approved):

PR #1600, opened 2026-05-22
fix(Components): shift in clipboard button [YTFRONT-5799]

SimbiozizV:
https://nda.ya.ru/t/32ZeaspK7d3d5X
  ## Summary by Sourcery

Adjust ClipboardButton to correctly handle copying alternative text when the Shift key is used and align UI package with a specific @ytsaurus/components build.

New Features:
- Allow ClipboardButton to provide a separate shift-activated text variant via a dedicated shiftText configuration in Storybook.

Bug Fixes:
- Fix ClipboardButton so that Shift key state is captured correctly and the appropriate text is copied and reported via onCopy callbacks.

Enhancements:
- Refine ClipboardButton integration with CopyToClipboard by delegating text resolution and copy callbacks to component methods.

Build:
- Pin @ytsaurus/components dependency in the UI package to a specific tarball URL instead of a version range.


packages/components/src/components/ClipboardButton/ClipboardButton.stories.tsx
```diff
@@ -8,6 +8,7 @@ const meta = {
     tags: ['autodocs'],
     args: {
         text: 'Some text',
+        shiftText: 'Some shift',
```
xexys: suggestion - `Some shift text`

packages/components/src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -81,24 +81,34 @@ export class ClipboardButton extends Component<ClipboardButtonProps> {
         );
     }
 
-    onClick = (event: React.MouseEvent<HTMLElement, MouseEvent>) => {
-        const {text, shiftText, onCopy} = this.props;
+    getTextToCopy = () => {
+        const {text, shiftText} = this.props;
```
xexys, line 85: Why text prop is optional?

xexys (dismissed):

xexys (dismissed):

xexys (approved):

PR #1604, opened 2026-05-25
feat(Navigation/Table): add rls-access meta table item [YTFRONT-5782]

ma-efremoff:
https://nda.ya.ru/t/DX_to51y7dd3YA
    ## Summary by Sourcery

Add navigation table metadata indicator for row-level access restrictions and wire it to permission checks via new YT checkPermissions API endpoint.

New Features:
- Display an 'Access' meta item in table metadata when row-level security limits a user's visible rows.
- Introduce a reusable hook to check table row-level security and surface corresponding status in the UI.

Enhancements:
- Add a YT checkPermissions API endpoint and React hook wrapper to query user access levels with auto-refresh support.
- Reduce CI Playwright workers on CI to lower parallelism for UI tests.

### DIFF
--- a/packages/ui/src/ui/pages/navigation/content/Table/table-hooks/useTableAccessMetaItem.tsx
+++ b/packages/ui/src/ui/pages/navigation/content/Table/table-hooks/useTableAccessMetaItem.tsx
@@ -1,0 +1,73 @@
+import {Lock} from '@gravity-ui/icons';
+import {Flex} from '@gravity-ui/uikit';
+import {Tooltip} from '@ytsaurus/components';
+import compact_ from 'lodash/compact';
+import React from 'react';
+import HelpLink from '../../../../../components/HelpLink/HelpLink';
+import Label from '../../../../../components/Label';
+import {YTErrorInline} from '../../../../../containers/YTErrorInline/YTErrorInline';
+import {YTApiId} from '../../../../../rum/rum-wrap-api';
+import {useCheckPermissionQuery} from '../../../../../store/api/yt/checkPermissions';
+import {useGetQuery} from '../../../../../store/api/yt/get';
+import UIFactory from '../../../../../UIFactory';
+
+type UseTableAccessMetaItemParams = {
+    path: string;
+    user: string;
+    transaction_id?: string;
+};
+
+export function useTableAccessMetaItem({path, user, transaction_id}: UseTableAccessMetaItemParams) {
+    const fullReadPermission = useCheckPermissionQuery({
+        id: YTApiId.checkPermissions,
+        parameters: {
+            path,
+            transaction_id,
+            permission: 'full_read',
+            user,
+        },
+    });
+    const hasRls = useGetQuery<boolean>({
+        id: YTApiId.checkPermissions,
+        parameters: {transaction_id, path: `${path}/@has_row_level_ace`},
+    });
+
+    if (fullReadPermission.error || hasRls.error) {
+        const error = {
+            message: 'Failed to check row level access.',
+            inner_errors: compact_([fullReadPermission.error, hasRls.error]),
+        };
+        return {
+            key: 'access',
+            label: 'Access',
+            value: <YTErrorInline error={error} />,
+        };
+    }
+
+    const hasLimitedAccess = hasRls.data && fullReadPermission.data?.action === 'deny';
+    if (!hasLimitedAccess) {
+        return undefined;
+    }
+
+    return {
+        key: 'access',
+        label: 'Access',
+        value: (
+            <Label theme="warning">
+                <Flex gap={1} alignItems="center">
+                    <Lock />
+                    <Tooltip
+                        content={
+                            <>
+                                You can see subset of rows according to your subset of permissions.{' '}
+                                <HelpLink url={UIFactory.docsUrls['acl:row-level-security']} />,
+                            </>
+                        }
+                    >
+                        Limited
+                    </Tooltip>
+                </Flex>
+            </Label>
+        ),
+    };
+}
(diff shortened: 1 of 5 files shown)


packages/ui/src/ui/pages/navigation/content/Table/table-hooks/useTableAccessMetaItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export function useTableAccessMetaItem({
+    path,
+    user,
+    transaction_id,
+}: {
+    path: string;
+    user: string;
+    transaction_id?: string;
+}) {
```
xexys: It's better to put it in a separate `Params` parameters type

packages/ui/src/ui/pages/navigation/content/Table/table-hooks/useTableAccessMetaItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+            inner_errors: compact_([fullReadPermission.error, hasRls.error]),
+        };
+        return {
+            key: 'access',
+            label: 'Access',
+            value: <YTErrorInline error={error} />,
+        };
+    }
+
+    const limitedAccess = hasRls.data && fullReadPermission.data?.action === 'deny';
```
xexys: `hasLimitedAccess` ?

packages/ui/src/ui/store/api/yt/checkPermissions/endpoint.ts
```diff
@@ -0,0 +1,19 @@
+import {type YTError} from '../../../../../@types/types';
+import {type YTApiIdType} from '../../../../../shared/constants/yt-api-id';
+import {type CheckPermissionsParams} from '../../../../../shared/yt-types';
+import {ytApiV3Id} from '../../../../rum/rum-wrap-api';
+import {type YTEndpointApiArgs} from '../types';
+
+export async function checkPermission({
+    id,
+    ...args
```
xexys: `args` is usually the name for the argument list, but in this case, it's not.

`params` or `rest` would be a better choice for the variable name in this case.

packages/ui/src/ui/store/api/yt/checkPermissions/endpoint.ts
```diff
@@ -0,0 +1,19 @@
...
+}: YTEndpointApiArgs<CheckPermissionsParams> & {
+    id: YTApiIdType;
+}) {
```
xexys: I propose to move it to a separate type

packages/ui/src/ui/store/api/yt/checkPermissions/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import {getUseAutoRefresh} from '../../../../store/selectors/settings/settings-ts';
+import {type OverrideDataType} from '../types';
+import {ytApi} from '../ytApi';
+import {checkPermission} from './endpoint';
+
+export const checkPermissionApi = ytApi.injectEndpoints({
+    endpoints: (build) => ({
+        checkPermissions: build.query({
+            queryFn: checkPermission,
+            providesTags: (_result, _error, args) => [args.id],
```
xexys: Same as https://github.com/ytsaurus/ytsaurus-ui/pull/1604/changes#r3334472167

packages/ui/src/ui/store/api/yt/checkPermissions/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    const useAutoRefresh = useSelector(getUseAutoRefresh);
+    const cluster = useSelector(selectCluster);
+
+    const options = {
+        pollingInterval: useAutoRefresh ? DEFAULT_UPDATER_TIMEOUT : undefined,
+        skipPollingIfUnfocused: true,
+    };
+
+    const customArgs =
+        'setup' in args
```
xexys: Logic based on the presence of a key in an object doesn't seem reliable. Maybe we should also check the value?

packages/ui/src/ui/store/api/yt/checkPermissions/index.ts
```diff
@@ -0,0 +1,37 @@
+import {DEFAULT_UPDATER_TIMEOUT} from '../../../../hooks/use-updater';
+import {useSelector} from '../../../../store/redux-hooks';
+import {selectCluster} from '../../../../store/selectors/global/cluster';
+import {getUseAutoRefresh} from '../../../../store/selectors/settings/settings-ts';
+import {type OverrideDataType} from '../types';
+import {ytApi} from '../ytApi';
+import {checkPermission} from './endpoint';
+
+export const checkPermissionApi = ytApi.injectEndpoints({
```
xexys: Is export necessary in this case?

We may get into a situation where two functions are exported that are essentially identical in name, but with different behavior.

Both `useCheckPermissionQuery` and `checkPermissionApi.useCheckPermissionsQuery` can be used when importing a module

xexys (changes requested):

packages/ui/src/ui/store/api/yt/checkPermissions/index.ts
ma-efremoff: Type of `setup` field is checked by typescript compilier. Here we just need to check if  the field is present.

packages/ui/src/ui/store/api/yt/checkPermissions/index.ts
```diff
@@ -0,0 +1,37 @@
...
+const checkPermissionApi = ytApi.injectEndpoints({
+    endpoints: (build) => ({
+        checkPermissions: build.query({
+            queryFn: checkPermission,
+            providesTags: (_result, _error, params) => [params.id],
+        }),
+    }),
+});
+
+export function useCheckPermissionQuery(args: CheckPermissionParams) {
```
xexys: `params`

xexys (dismissed):

xexys (approved):

PR #1636, opened 2026-06-10
PrometheusDashboard fixes

ma-efremoff:
https://nda.ya.ru/t/FRKMYFur7euNhx

### DIFF
--- a/packages/ui/src/ui/containers/PrometheusDashboard/plugins/timeseries/timeseries-config.ts
+++ b/packages/ui/src/ui/containers/PrometheusDashboard/plugins/timeseries/timeseries-config.ts
@@ -1,24 +1,22 @@
 import React from 'react';
 import reduce_ from 'lodash/reduce';
+import cloneDeep_ from 'lodash/cloneDeep';
 import {type FieldConfig} from '../../../../../shared/prometheus/types';
 
 export type PrometheusChartFieldConfig = {
     showLegend?: boolean;
     axisLabel?: string;
-    propertiesByRefId: Record<string, PrometheusChartProperties>;
-};
-
-export type PrometheusChartProperties = {
-    unit?: 'bytes' | unknown;
+    propertiesByRefId: Record<string, FieldConfig['defaults']>;
 };
 
 export function usePrometheusChartFieldConfig(
     fieldConfig: FieldConfig,
 ): PrometheusChartFieldConfig {
     const res = React.useMemo(() => {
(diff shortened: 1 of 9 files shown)


packages/ui/src/ui/containers/PrometheusDashboard/plugins/timeseries/timeseries-config.ts
```diff
@@ -0,0 +1,59 @@
...
+
+                    return acc;
+                },
+                {} as Record<string, typeof defaults>,
+            ),
+        };
+    }, [fieldConfig]);
+
+    return res;
+}
```
xexys, line 49: This code is quite hard to read. 
Maybe we could make separate function for the internal reduce?

packages/ui/src/ui/containers/PrometheusDashboard/utils/prometheus-format.ts
```diff
@@ -0,0 +1,25 @@
...
+        case 'Bps':
+            return (v: number) => format.Bytes(v) + '/s';
+        case 'iops':
+            return (v: number) => format.NumberSmart(v) + 'io/s';
+        default:
+            return format.NumberSmart;
+    }
+}
+
+function formatGBytes(value: number | string | undefined) {
```
xexys: It would be great to get rid of the `undefined` value and just not call the function if there is no argument

packages/ui/src/ui/containers/PrometheusDashboard/utils/prometheus-format.ts
```diff
@@ -0,0 +1,25 @@
+import {type PrometheusDashboardUnitType} from '../../../../shared/prometheus/types';
+import format from '../../../common/hammer/format';
+
+export function getPrometheusFormatter(unit: PrometheusDashboardUnitType | unknown) {
```
xexys: This type looks strange because it somehow comes down to `unknown`.

It seems like there should always be a `PrometheusDashboardUnitType` here.

xexys (changes requested):

packages/ui/src/ui/containers/PrometheusDashboard/utils/prometheus-format.ts
ma-efremoff: It demonstrates there are might be other values we don't want to handle in a special way (also `unit` might be `undefined`). We prefer `format.NumberSmart` for rest cases.

xexys (approved):

PR #1658, opened 2026-06-19
fix(System): manual container selection [YTFRONT-5778]

SimbiozizV:
https://nda.ya.ru/t/eYMY4tn27frV33
 ## Summary by Sourcery

Replace the master container dropdown with a suggest-based input to improve manual container selection in the System Monitoring Prometheus page.

New Features:
- Allow users to manually select or filter master containers via a suggest input with text-based search.

Bug Fixes:
- Fix incorrect or unreliable manual container selection by normalizing and validating the selected container value.

Enhancements:
- Normalize container option handling by storing items as plain strings and centralizing container reset and apply logic.

### DIFF
--- a/packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
+++ b/packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
@@ -1,9 +1,11 @@
-import React from 'react';
+import React, {useCallback} from 'react';
 import sortBy_ from 'lodash/sortBy';
 
-import {Flex, Select} from '@gravity-ui/uikit';
+import {Flex} from '@gravity-ui/uikit';
+import Suggest, {type SuggestItem} from '../../../components/Suggest/Suggest';
 import {useFetchBatchQuery} from '../../../store/api/yt';
 import {YTApiId} from '../../../rum/rum-wrap-api';
+import {filterSuggestItems} from './helpers/filterSuggestItems';
 
 export function MasterLocalContainers({
     allValue,
@@ -20,23 +22,40 @@ export function MasterLocalContainers({
         errorTitle: 'Failed to load masters',
     });
 
-    const options = React.useMemo(() => {
-        const res = [{value: allValue, content: allValue}];
-        sortBy_(data?.[0]?.output).forEach((i) => {
-            res.push({value: i, content: i});
+    const items = React.useMemo(() => {
+        const res = [allValue];
+        sortBy_(data?.[0]?.output).forEach((item) => {
+            res.push(item);
         });
         return res;
     }, [data, allValue]);
 
+    const applyContainer = useCallback(
+        (item: SuggestItem) => {
+            const value = typeof item === 'string' ? item : item.value;
+            setContainer(value);
+        },
+        [setContainer],
+    );
+
+    const onEnterKeyDown = useCallback(
+        (value: string) => {
+            if (value.trim()) {
+                applyContainer(value);
+            }
+        },
+        [applyContainer],
+    );
+
     return (
         <Flex gap={1} alignItems="center">
             Container:
-            <Select
-                options={options}
-                value={[container]}
-                onUpdate={([v]) => {
-                    setContainer(v);
-                }}
+            <Suggest
+                text={container}
+                items={items}
+                filter={filterSuggestItems}
+                apply={applyContainer}
+                onEnterKeyDown={onEnterKeyDown}
             />
         </Flex>
     );
--- a/packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/helpers/filterSuggestItems.ts
+++ b/packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/helpers/filterSuggestItems.ts
@@ -1,0 +1,11 @@
+import type {SuggestItem} from '../../../../components/Suggest/Suggest';
+
+export const filterSuggestItems = (list: Array<SuggestItem>, text?: string) => {
+    if (!text) return list;
+
+    const lcText = text.toLowerCase();
+    return list.filter((item) => {
+        const itemText = typeof item === 'string' ? item : item.value;
+        return itemText.toLowerCase().includes(lcText);
+    });
+};


packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
```diff
@@ -20,23 +21,50 @@ export function MasterLocalContainers({
         errorTitle: 'Failed to load masters',
     });
 
-    const options = React.useMemo(() => {
-        const res = [{value: allValue, content: allValue}];
+    const items = React.useMemo(() => {
+        const res = [allValue];
         sortBy_(data?.[0]?.output).forEach((i) => {
-            res.push({value: i, content: i});
+            res.push(i);
```
xexys: I suggest not to use abbreviations and call `item`

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
```diff
@@ -20,23 +21,50 @@ export function MasterLocalContainers({
...
+        });
+    }, []);
+
+    const applyContainer = React.useCallback(
+        (value: SuggestItem) => {
+            const v = typeof value === 'string' ? value : value.value;
+            setContainer(v);
+        },
+        [setContainer, allValue],
+    );
```
xexys, line 39: I suggest not to use abbreviations

```ts
const applyContainer = React.useCallback(
    (item: SuggestItem) => {
        const value = typeof item === 'string' ? item : item.value;
        setContainer(value);
    },
    [setContainer, allValue],
);
```

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
```diff
@@ -20,23 +21,50 @@ export function MasterLocalContainers({
...
+            return itemText.toLowerCase().includes(lcText);
+        });
+    }, []);
+
+    const applyContainer = React.useCallback(
+        (value: SuggestItem) => {
+            const v = typeof value === 'string' ? value : value.value;
+            setContainer(v);
+        },
+        [setContainer, allValue],
```
xexys: `allValue` - unnecessary dependence

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
```diff
@@ -20,23 +21,50 @@ export function MasterLocalContainers({
...
+    );
+
+    const onEnterKeyDown = React.useCallback(
```
xexys: Suggestion - use `useCallback` instead of `React.useCallback`, because there is less noise in the code

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
```diff
@@ -20,23 +21,50 @@ export function MasterLocalContainers({
...
+        (value: string) => {
+            if (value.trim()) {
+                applyContainer(value);
+            }
+        },
+        [applyContainer],
+    );
```
xexys, line 48: Could you explain the behavior of this function? Why is `applyContainer` only called for empty values?

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
```diff
@@ -20,23 +21,50 @@ export function MasterLocalContainers({
...
+    const items = React.useMemo(() => {
+        const res = [allValue];
         sortBy_(data?.[0]?.output).forEach((i) => {
-            res.push({value: i, content: i});
+            res.push(i);
         });
         return res;
     }, [data, allValue]);
 
+    const filterItems = React.useCallback((list: Array<SuggestItem>, text?: string) => {
```
xexys: It looks like the correct type is `text: string | undefined` ?

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
```diff
@@ -20,23 +21,50 @@ export function MasterLocalContainers({
...
+        if (!text) return list;
+
+        const lcText = text.toLowerCase();
+        return list.filter((item) => {
+            const itemText = typeof item === 'string' ? item : item.value;
+            return itemText.toLowerCase().includes(lcText);
+        });
+    }, []);
```
xexys: `filterItems` function can be moved outside the component

xexys (changes requested):

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
SimbiozizV: It seems that you are a little confused about the condition. It is only called for non-empty values

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/MasterLocalContainers.tsx
SimbiozizV: An equivalent type is currently applied to it

packages/ui/src/ui/pages/system/SystemMonitoringPrometheus/helpers/filterSuggestItems.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const filterSuggestItems = (list: Array<SuggestItem>, text?: string) => {
+    if (!text) return list;
+
+    const lcText = text.toLowerCase();
+    return list.filter((item) => {
+        const itemText = typeof item === 'string' ? item : item.value;
+        return itemText.toLowerCase().includes(lcText);
+    });
+};
```
xexys, line 11: We use `utils`, not `helpers` :)

xexys (changes requested):

xexys (approved):

PR #1664, opened 2026-06-23
chore(Pages): add i18n [YTFRONT-5069]

ma-efremoff:
https://nda.ya.ru/t/mL6Eq-m67gHbBn
 ## Summary by Sourcery

Introduce an i18n-enabled app pages registry and wire it into navigation and menu state.

New Features:
- Add a centralized APP_PAGES registry with i18n-based titles and support for registering extra pages at runtime.

Enhancements:
- Update slideout menu state to consume the new APP_PAGES registry instead of the legacy pages list.
- Adjust AppNavigation layout styles to make the side panel fill available height and simplify the pages editor list virtualization settings.

### DIFF
--- a/packages/ui/src/ui/containers/AppNavigation/PagesEditorPanel.tsx
+++ b/packages/ui/src/ui/containers/AppNavigation/PagesEditorPanel.tsx
@@ -44,7 +44,6 @@ export function PagesEditorPanel() {
             renderItem={(item) => <PagesEditorListItem {...item} onClick={handleClick} />}
             itemHeight={40}
             onSortEnd={handleSort}
-            virtualized
         />
     );
 }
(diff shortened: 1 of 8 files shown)


packages/ui/src/ui/containers/AppNavigation/PagesEditorPanel.tsx
```diff
@@ -44,7 +44,6 @@ export function PagesEditorPanel() {
             renderItem={(item) => <PagesEditorListItem {...item} onClick={handleClick} />}
             itemHeight={40}
             onSortEnd={handleSort}
-            virtualized
```
SimbiozizV, line 47: Is it normal that virtualization is now disabled in the list?

ma-efremoff: Yes we don't need virtualization here case amount of elements is always small

SimbiozizV (approved):

PR #1670, opened 2026-06-25
fix(Queries): dont add discovery path in spark connect [YTFRONT-5887]

SimbiozizV:
https://nda.ya.ru/t/VKGlOO2T7gUm9u
## Summary by Sourcery

Bug Fixes:
- Prevent assigning a discovery path for SPYT queries when SPYT Connect is available, avoiding incorrect discovery group usage in Spark Connect mode.

### DIFF
--- a/packages/ui/src/ui/store/actions/query-tracker/query.ts
+++ b/packages/ui/src/ui/store/actions/query-tracker/query.ts
@@ -84,7 +84,10 @@ import {
     resetQueryTabs,
 } from '../../reducers/query-tracker/queryTabsSlice';
 import {updateQueryTabs} from './queryTabs/queryTabs';
-import {selectSpytDefaultSettings} from '../../selectors/query-tracker/queryTrackerEnginesInfo';
+import {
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/store/actions/query-tracker/query.ts
```diff
@@ -127,14 +130,15 @@ export const setUserLastChoice =
         const lastClique = selectLastUserChoiceQueryChytClique(state);
         const lastVersion = selectLastUserChoiceYqlVersion(state);
         const defaultYqlVersion = selectDefaultYqlVersion(state);
+        const hasSpytConnect = selectAvailableSpytConnect(state);
```
xexys: The variable name `hasSpytConnect` and the selector name `selectAvailableSpytConnect` have slightly different meanings.

Maybe something needs to be renamed?

xexys (changes requested):

xexys (approved):

PR #1681, opened 2026-07-01
fix(UI): format min/max bounds in NumberInput validation errors [YTFR…

SimbiozizV:
https://nda.ya.ru/t/RTkDzKc67h9KMw
 …ONT-5452]

## Summary by Sourcery

Bug Fixes:
- Ensure min/max bounds in NumberInput validation errors respect custom format functions and decimal formatting settings.

### DIFF
--- a/packages/ui/src/ui/components/NumberInput/NumberInput.tsx
+++ b/packages/ui/src/ui/components/NumberInput/NumberInput.tsx
@@ -37,6 +37,20 @@ export function formatValue(
     return res === hammer.format.NO_VALUE ? '' : res;
 }
 
+function formatBoundForError(bound: number, props: NumberInputWithErrorProps) {
+    const {format, decimalPlaces, formatFn} = props;
+    const formatted = formatFn
+        ? formatFn(bound)
+        : formatValue(bound, format, {digits: decimalPlaces});
+    const raw = String(bound);
+
+    if (!formatted || formatted === raw) {
+        return raw;
+    }
+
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/components/NumberInput/NumberInput.tsx
```diff
@@ -116,6 +116,14 @@ export class NumberInputWithError extends React.Component<NumberInputWithErrorPr
         return null;
     }
 
+    static formatBound(bound: number, props: NumberInputWithErrorProps) {
```
ma-efremoff: It is better to avoid static members

packages/ui/src/ui/components/NumberInput/NumberInput.tsx
```diff
@@ -146,11 +154,15 @@ export class NumberInputWithError extends React.Component<NumberInputWithErrorPr
         }
 
         if (min !== undefined && value < min) {
-            return i18n('error_must-be-gte', {min});
+            return i18n('error_must-be-gte', {
+                min: NumberInputWithError.formatBound(min, props),
```
ma-efremoff: In some cases, users may need to copy the exact value. Therefore, it is better to save such value without formatting. It is better to display the formatted value in parentheses:
```
1536 (1.5KB)
```

xexys (approved):

ma-efremoff (approved):

PR #1691, opened 2026-07-08
feat(Scheduling): show active operation ref [YTFRONT-5666]

SimbiozizV:
https://nda.ya.ru/t/go2vkwxG7htxRn
     ## Summary by Sourcery

Show the currently pinned operation in the scheduling overview toolbar and allow clearing it.

New Features:
- Display a badge in the scheduling overview toolbar indicating the active (pinned) operation reference with contextual text.
- Provide a close action on the active operation badge to clear the pinned operation reference from scheduling state.

Enhancements:
- Update scheduling overview meta labels for mode and weight to use more specific i18n keys.
- Introduce a dedicated i18n setup for the scheduling toolbar component and re-export it from a local index file.


packages/ui/src/ui/pages/scheduling/Content/tabs/Overview/SchedulingToolbar/i18n/en.json
```diff
@@ -0,0 +1,17 @@
+{
+  "context_pinned-operation": "Pinned operation: {{operationRefId}}",
```
ma-efremoff: I am not sure `Pinned` word is the best one here. It is not clear what does it means here, but if we want to keep it we should add tooltip that clarify the scrolling behaviour. 
It will be more cleare to use `Scroll to operation` or `Scroll to` here

ma-efremoff (approved):

PR #1708, opened 2026-07-20
fix(Queries): utf8 result problem [YTFRONT-5937]

SimbiozizV:
https://nda.ya.ru/t/XjG-Q7Yz7jPWR2
## Summary by Sourcery

Adjust query result UTF-8 decoding behavior and add coverage for Cyrillic YSON query results.

New Features:
- Add an end-to-end query tracker test that verifies Cyrillic strings in YSON query results are rendered correctly.

Bug Fixes:
- Prevent double UTF-8 decoding of query results by disabling decodeUTF8 in result views and cell preview for queries.

Enhancements:
- Allow CellPreviewModal to accept overridden Unipika/YSON settings, including optional decodeUTF8.
- Extend YsonSettings selector typing to include an optional decodeUTF8 flag.

Tests:
- Extend e2e cluster initialization to create a YSON table with Cyrillic data and persist its query id for UI tests.
- Expose E2E_YSON_CYRILLIC_QUERY_ID in test utilities and increase query result table visibility timeout for flaky UI tests.

Chores:
- Add test data files for the YSON Cyrillic table used in e2e tests.


packages/ui/src/ui/containers/CellPreviewModal/CellPreviewModal.tsx
```diff
@@ -127,7 +127,11 @@ function PreviewContent(props: PreviewContentProps) {
                     virtualized
                     value={getUnipikaNodeForYsonPreview(data)}
                     scrollContainer={scrollContainer}
-                    settings={unipikaSettings}
+                    settings={
+                        data?.$type === 'yql.yson'
+                            ? {...unipikaSettings, decodeUTF8: false}
+                            : unipikaSettings
+                    }
```
xexys: It seems strange to me that the calling code passes `data` and `unipikaSettings`, but `PreviewContent` overrides the settings.

 It should be either
1. Pass the correct `unipikaSettings` in the parent component.

2. The `PreviewContent` component only accepts `data` and uses the `data` to select the rendering settings.

What do you think?

xexys (changes requested):

xexys (approved):

PR #1709, opened 2026-07-20
chore(Settings): some fixes for tests [YTFRONT-5069]

ma-efremoff:
https://nda.ya.ru/t/MzhhbNvh7jBLmv

### DIFF
--- a/packages/ui/src/ui/containers/SettingsPanel/SettingsPanel.tsx
+++ b/packages/ui/src/ui/containers/SettingsPanel/SettingsPanel.tsx
@@ -23,7 +23,12 @@ function SettingsPanel() {
                     const hasItems = some_(page.sections, ({items}) => items.length > 0);
                     return (
                         hasItems && (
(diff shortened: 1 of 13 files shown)


packages/ui/src/ui/containers/SettingsPanel/SettingsPanel.tsx
```diff
@@ -23,7 +23,12 @@ function SettingsPanel() {
                     const hasItems = some_(page.sections, ({items}) => items.length > 0);
                     return (
                         hasItems && (
-                            <Settings.Page key={i} title={page.title} icon={iconData(page.icon)}>
+                            <Settings.Page
+                                key={i}
```
xexys: Usually, `index` is a bad choice for key. Is it possible to use `page.id` here?

xexys (changes requested):

xexys (approved):

PR #1719, opened 2026-07-29
fix(Operation): wrong suspended status [YTFRONT-5948]

SimbiozizV:
https://nda.ya.ru/t/rJ7u-5rB7kAEAp
## Summary by Sourcery

Fix operation detail status handling for suspended and GPU operations so that waiting states and labels are shown correctly.

Bug Fixes:
- Correct special waiting status calculation to ignore suspended operations and properly handle GPU vanilla operations.
- Ensure main status label and running icon state are consistent when operations are waiting for jobs or resources.

### DIFF
--- a/packages/ui/src/ui/pages/operations/OperationDetail/OperationDetail.tsx
+++ b/packages/ui/src/ui/pages/operations/OperationDetail/OperationDetail.tsx
@@ -106,7 +106,7 @@ function getSpecialWaitingStatuses(
     type: string | undefined,
     isGpuOperation: boolean | undefined,
 ): {isWaitingForJobs?: boolean; isWaitingForResources?: boolean} {
-    if (state !== 'running') {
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui/pages/operations/OperationDetail/OperationDetail.tsx
```diff
@@ -118,7 +118,7 @@ function getSpecialWaitingStatuses(
 
     switch (type) {
         case 'vanilla': {
-            const isSpecialStatus = state === 'running' && isGpuOperation && !suspended;
+            const isSpecialStatus = isGpuOperation;
```
xexys: The `isSpecialStatus` variable looks redundant. Should it be removed?

xexys (changes requested):

xexys (approved):

PR #1736, opened 2026-08-04
feat(MetaTable): render nirvana block url and fixes [YTFRONT-5942]

xexys:
## Summary by Sourcery

Render Nirvana block URL metadata in MetaTable and introduce reusable link, ID, readable value, and formatted value templates while tightening MetaTable config typing.

New Features:
- Display Nirvana block URL in MetaTable with either markdown rendering or a link component supporting truncation and copy-to-clipboard.

Bug Fixes:
- Correct QA attribute naming for TTL expiration path fields and align element keys with their semantics.

Enhancements:
- Refactor MetaTable templates into dedicated components for IDs, links, readable values, and formatted values with consistent styling and clipboard support.
- Extend navigation meta configuration with an optional markdown renderer and use a strongly-typed, partial-based config shape across MetaTable presets.
- Improve TTL and size presets to use the new link and formatted value templates, including ellipsis, max-width constraints, and clipboard buttons.
- Simplify MetaTable SCSS by removing layout rules now handled by new template components.


xexys:
@ma-efremoff @SimbiozizV 

The tests were performed in a different PR
Screen tests fail there—that's normal. They will be fixed when the library version is updated.
https://github.com/ytsaurus/ytsaurus-ui/pull/1735

packages/components/src/components/MetaTable/presets/ttl.tsx
```diff
@@ -17,6 +17,8 @@ import './ttl.scss';
 
 const block = cn('meta-table-ttl');
 
+const LINK_MAX_WIDTH = '50ch';
```
SimbiozizV: I’m confused because this variable is defined in two places. Here and in /MetaTable/presets/main.tsx

If we want the same behavior in the presets, maybe we can move it somewhere higher up and reuse it?

packages/components/src/components/MetaTable/presets/ttl.tsx
```diff
@@ -82,7 +93,12 @@ export function makeTTLItems(
             label: i18n('field_effective-expiration-timeout-path'),
             qa: 'expiration_timeout_path',
             value: timeoutPathUrl ? (
-                <Template.Link url={timeoutPathUrl} text={timeout.path} />
+                <TemplateLink
+                    url={timeoutPathUrl}
+                    text={timeout.path}
+                    maxWidth={LINK_MAX_WIDTH}
```
SimbiozizV, line 98: As an option, you can set 50 as the default value for this field. Or are there other sizes somewhere?

packages/components/src/components/MetaTable/templates/TemplateFormattedValue.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import {itemBlock} from './utils';
+
+type Props = {
+    value?: string | number;
+    format?: string | ((value: unknown, settings?: Record<string, unknown>) => ReactNode);
+    settings?: Record<string, unknown>;
+};
+
+export function TemplateFormattedValue({value, format: formatKey, settings}: Props) {
```
SimbiozizV: I’m concerned that in `Props` this parameter is defined as `format`, but then it’s redefined as `formatKey`.

`Props` isn’t exported anywhere. Maybe we should change the key from `format` to `formatKey` so we don’t have to redefine it later?

packages/components/src/components/MetaTable/templates/TemplateFormattedValue.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const fmtClass = fmtIsFunc ? undefined : (formatKey as string | undefined)?.toLowerCase();
+    return (
+        <span className={itemBlock('value', {format: fmtClass})}>
+            {fmtIsFunc
+                ? (formatKey as (value: unknown, settings?: Record<string, unknown>) => ReactNode)(
+                      value,
+                      settings,
+                  )
+                : formatKey
+                  ? hammerFormat[formatKey as keyof typeof hammerFormat](value, settings)
```
SimbiozizV: double ternary operator

packages/components/src/components/MetaTable/templates/TemplateId.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+type Props = {
+    id?: string;
+};
+
+export function TemplateId({id}: Props) {
+    return (
+        <Flex gap={1} wrap="nowrap" alignItems="center" className={itemBlock('id')}>
+            <Text ellipsis>{id}</Text>
+            <ClipboardButton view="flat-secondary" text={id ?? ''} size="s" />
```
SimbiozizV: Should this component even be displayed if `id === undefined`?

packages/components/src/components/MetaTable/templates/TemplateLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+            style={{maxWidth}}
+        >
+            <Text ellipsis>
+                <Link title={url} href={url}>
+                    {IconComponent && <Icon data={IconComponent} size={14} />}
+                    {text}
+                </Link>
+            </Text>
+
+            {withClipboard && (
```
SimbiozizV, line 35: What will this button copy if `text` and `shiftText` are equal to `undefined`?

packages/components/src/components/MetaTable/templates/utils.ts
```diff
@@ -0,0 +1,3 @@
+import cn from 'bem-cn-lite';
+
+export const itemBlock: ReturnType<typeof cn> = cn('meta-table-item');
```
SimbiozizV: From the name, it’s not entirely clear what it is.

packages/components/src/components/MetaTable/templates/TemplateLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+    hoverContent,
+    maxWidth,
+}: Props) {
+    return (
+        <Flex
+            gap={1}
+            wrap="nowrap"
+            alignItems="center"
+            className={itemBlock('link')}
+            style={{maxWidth}}
```
SimbiozizV: Can’t we set this property directly in `meta-table-item__link`?

packages/components/src/components/MetaTable/presets/ttl.tsx
xexys: I created the variable separately because these are different presets with their own settings, depending on the dataset they contain.

packages/components/src/components/MetaTable/presets/ttl.tsx
xexys: I’m not sure this is correct. 

The `TemplateLink` component was internal until today, but now it’s part of the library’s interface, and it seems like it wouldn’t be very right to hard‑code a length limitation into it.

packages/components/src/components/MetaTable/templates/TemplateFormattedValue.tsx
xexys: This is just a copy of the original code with no changes.

https://github.com/ytsaurus/ytsaurus-ui/blob/fe5de5c62a6e06e0adaa0a7746c5d61874ef0909/packages/components/src/components/MetaTable/templates/Template.tsx#L27

But I can fix this.

packages/components/src/components/MetaTable/templates/TemplateFormattedValue.tsx
xexys: Yep, it’s a copy too. I can fix it.

packages/components/src/components/MetaTable/templates/TemplateId.tsx
xexys: That’s a good question; I think, of course, not. 
But this contract already existed, and I didn’t want to break the public interface.

https://github.com/ytsaurus/ytsaurus-ui/blob/fe5de5c62a6e06e0adaa0a7746c5d61874ef0909/packages/components/src/components/MetaTable/templates/Template.tsx#L15

Let’s make it mandatory, but we won’t release the major version.

packages/components/src/components/MetaTable/templates/TemplateLink.tsx
xexys: It will copy empty text, as it was in original code
https://github.com/ytsaurus/ytsaurus-ui/blob/fe5de5c62a6e06e0adaa0a7746c5d61874ef0909/packages/components/src/components/MetaTable/templates/Template.tsx#L67-L68

packages/components/src/components/MetaTable/templates/utils.ts
xexys: You’re suggesting to call it `metaTableItemBlock`?

packages/components/src/components/MetaTable/templates/TemplateLink.tsx
xexys: `maxWidth` is a dynamic property. 

You’ll have to prepare a mapping from class names, or do it via a CSS variable, which seems like overkill.

packages/components/src/components/MetaTable/templates/TemplateLink.tsx
SimbiozizV: I’m more talking about whether it’s necessary to display such a button if it’s essentially not functional in this case

packages/components/src/components/MetaTable/templates/utils.ts
SimbiozizV: as an option

SimbiozizV (dismissed):

SimbiozizV (approved):

PR #1752, opened 2026-08-13
feat(Flow): add pipeline state viewer [#1752]

blinkov:
https://nda.ya.ru/t/YrnN8Jq37nQQiQ


packages/ui/src/ui/constants/index.ts
```diff
@@ -41,6 +41,14 @@ export const TYPED_OUTPUT_FORMAT = {
     },
 };
 
+export const TYPED_INPUT_FORMAT = {
```
ma-efremoff: There is no need to copy-paste TYPED_OUTPUT_FORMAT
```ts
export const TYPED_INPUT_FORMAT = TYPED_OUTPUT_FORMAT;
```

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
```diff
@@ -0,0 +1,227 @@
...
+    | 'validation_fill-all-keys'
+    | 'validation_key-target-mismatch'
+    | 'validation_no-scope';
+
+export type FlowStateValidationError = {
+    errorKey: FlowStateValidationErrorKey;
+    params?: Record<string, string>;
+};
+
+export type FlowAnnotatedInteger = {$type: 'int64' | 'uint64'; $value: string};
```
ma-efremoff: It is better to avoid direct access to $type, $value, $attributes. We prefer to use `ypath.getValue`

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.tsx
```diff
@@ -0,0 +1,147 @@
+import React from 'react';
+import cn from 'bem-cn-lite';
+
+import {Flex, Table, type TableColumnConfig, Text} from '@gravity-ui/uikit';
```
ma-efremoff: We prefer to use src/components/DataTableGravity instead of Table from uikit.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+import CollapsibleSection from '../../../../../components/CollapsibleSection/CollapsibleSection';
+import Link from '../../../../../containers/Link/Link';
```
ma-efremoff: Link is marked as depricated

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.tsx
```diff
@@ -0,0 +1,147 @@
...
+export type FlowComputationMessagesProps = {
+    path: string;
+    computation: string;
+    messages?: Array<FlowMessageType>;
+};
+
+function useStaticSpec(pipelinePath: string): FlowStaticSpec | undefined {
+    const [staticSpec, setStaticSpec] = React.useState<FlowStaticSpec>();
+
+    React.useEffect(() => {
```
ma-efremoff: We prefer rtk-query endpoints instead of useEffect, see src/store/api/flow

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
```diff
@@ -0,0 +1,56 @@
...
+export function flowDeleteStates(pipeline_path: string, body: FlowDeleteStatesBody) {
+    return flowStateExecute<FlowDeleteStatesResponse>('delete-states', pipeline_path, body, {
+        input_format: TYPED_INPUT_FORMAT,
+    });
+}
+
+export function fetchPipelineState(pipeline_path: string) {
+    return ytApiV4.getPipelineState({
+        parameters: {pipeline_path},
+    }) as Promise<FlowPipelineStateValue>;
```
ma-efremoff: We prefer to avoid such cast `as ...`

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
ma-efremoff: It is better to avoid direct read of $type, $value fields. We prefer to use ypath.getValue, ypath.getNumber

ma-efremoff: It is better to avoid direct access to `$type`/`$value`/`$attributes` fields. We prefer to use `ypath` get such values

(here and bellow)

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
```diff
@@ -0,0 +1,227 @@
+export type FlowStateTarget = 'all' | 'key_state' | 'partition_state' | 'external_key_state';
```
ma-efremoff: There are some duplicates of types from  src/shared/yt-types

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.spec.tsx
```diff
@@ -0,0 +1,140 @@
+/** @jest-environment jsdom */
```
ma-efremoff: The tests in the file don't looks really useful. It is better to remove all *.spec.tsx files. If you really want to add some component's tests it is better to add storybook-ones with screenshots.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.tsx
ma-efremoff: We prefer src/components/DataTableGravity instead of Table from uikit.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowDeleteStatesDialog/FlowDeleteStatesDialog.tsx
```diff
@@ -0,0 +1,238 @@
...
+                    {(response?.errors ?? []).map((message, index) => (
+                        <YTErrorBlock key={`${index}:${message}`} error={{message}} />
+                    ))}
+                </React.Fragment>
+            ))}
+        </Flex>
+    );
+}
+
+export function FlowDeleteStatesDialog({
```
ma-efremoff, line 30: It is better to split component. 
For edit/confirmation actions we prefer to use YTDFDialog.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateSection/FlowStateSection.tsx
```diff
@@ -0,0 +1,283 @@
...
+function useDebouncedValue<T>(value: T, delayMs: number): T {
+    const [debounced, setDebounced] = React.useState(value);
+    React.useEffect(() => {
+        const timer = window.setTimeout(() => setDebounced(value), delayMs);
+        return () => window.clearTimeout(timer);
+    }, [value, delayMs]);
+    return debounced;
+}
+
+export function FlowStateSection({
```
ma-efremoff, line 34: It is better to split the component for smaller ones

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateSection/FlowStateSection.tsx
```diff
@@ -0,0 +1,283 @@
...
+    };
+
+    const handleDeleteCommitted = () => {
+        setRowSelection({});
+        runRead(filters, false);
+    };
+
+    return (
+        <div className={block()}>
+            <FlowStateFilters
```
ma-efremoff: We prefer to use `WithStickyToolbar` for panels with filters

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
```diff
@@ -0,0 +1,56 @@
+import {TYPED_INPUT_FORMAT, TYPED_OUTPUT_FORMAT} from '../../../../constants';
```
ma-efremoff: It is better to use rtk-query endpoints similar to existing Flow-components

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.ts
```diff
@@ -0,0 +1,1074 @@
...
+    const {value: path, attributes} = unwrapYsonNode(pathNode);
+    if (typeof path !== 'string' || !path.length) {
+        return undefined;
+    }
+    const cluster = attributes['cluster'];
+    return {path, cluster: typeof cluster === 'string' ? cluster : undefined};
+}
+
+const HEAVY_HITTERS_TITLE_PATTERN = /^Top \d+ heavy hitters$/;
+const HEAVY_HITTER_ENTRY_PATTERN = /^Key=([\s\S]*), Ratio=([^,]+), PartitionId=([^,]+)$/;
```
ma-efremoff: Such regexps looks unstable. It is better to have specific fields in API response.

packages/ui/src/ui/constants/index.ts
ma-efremoff: Why do we need it?

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.ts
```diff
@@ -0,0 +1,1074 @@
+import type {
```
ma-efremoff: The file is too big.
It looks like the file contains some parsing helpers that used from components's code. It is better do such transformations on requests-level.

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
```diff
@@ -24,6 +28,18 @@ import i18n from './i18n';
...
 
+type FlowComputationExtraTab = {
+    value: string;
+    title: string;
+    component: React.ComponentType<FlowComputationMonitorProps>;
+};
+
+const isLocalMode = Boolean(YT.isLocalCluster);
+
+const extraTabs: Array<FlowComputationExtraTab> = isLocalMode
```
ma-efremoff: Why do we need the check?

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
```diff
@@ -87,6 +133,12 @@ function FlowComputationTabs({computation}: {computation: string}) {
                     url: `/${cluster}/${Page.FLOWS}/computations/${encodeURIComponent(computation)}/monitor`,
```
ma-efremoff: We prefer to use corresponding helpers from `src/ui/utils/app-url`

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
```diff
@@ -114,7 +166,15 @@ function FlowComputationDetails({computation}: {computation: string}) {
             {Boolean(error) && <FlowError error={error} />}
             <FlowComputationPerformance data={data} onClick={onClick} />
             <div className={block('messages')}>
-                <FlowMessagesCollapsible messages={data?.messages} />
+                {isLocalMode ? (
```
ma-efremoff: Why do we need the check?

packages/ui/src/ui/pages/flow/Flow/Flow.tsx
```diff
@@ -40,6 +43,18 @@ import {getFlowPathMetaItems} from '../flow-components/FlowMeta/FlowMeta';
 
 const block = cn('yt-flow');
 
+type FlowPipelineExtraTab = {
+    value: string;
+    title: string;
+    component: React.ComponentType<{pipeline_path: string}>;
+};
+
+const isLocalMode = Boolean(YT.isLocalCluster);
```
ma-efremoff: ?

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
```diff
@@ -0,0 +1,227 @@
...
+export type FlowStateResultRow = {
+    section: FlowStateResultSection;
+    computationId?: string;
+    partitionId?: string;
+    key?: unknown;
+    stateName: string;
+    value: unknown;
+};
+
+export type FlowPipelineStateValue =
```
ma-efremoff: At least this type has duplicate in src/shared/yt-types. What else types are duplicated?

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
ma-efremoff: It is better to avoid such type-casts

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function fetchPipelineState(pipeline_path: string) {
+    return ytApiV4.getPipelineState({
+        parameters: {pipeline_path},
+    }) as Promise<FlowPipelineStateValue>;
+}
+
+export function fetchSpec(pipeline_path: string) {
+    return ytApiV4.getPipelineSpec({parameters: {pipeline_path}}) as Promise<FlowPipelineSpecData>;
```
ma-efremoff: same here

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
ma-efremoff: We prefer to use rtk-query endpoints, see src/ui/store/yt/flow. Such endpoints should return data ready to work, it should not contain fields like $value/$type.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
```diff
@@ -0,0 +1,56 @@
...
+    return rawFlowExecute(`${YTApiId.flowExecute}_${flowCommand}`, {
+        parameters: {flow_command: flowCommand, pipeline_path, ...extraParameters},
+        data: body,
+    }) as Promise<TResponse>;
+}
+
+export function flowReadStates(pipeline_path: string, body: FlowReadStatesBody) {
+    return flowStateExecute<unknown>('read-states', pipeline_path, body, {
+        input_format: TYPED_INPUT_FORMAT,
+        output_format: TYPED_OUTPUT_FORMAT,
```
ma-efremoff: As I can see in rest commits these parameters bring some unnecessary parsing values code. We need to discuss it.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
ma-efremoff: Some types in the file contains `$type`, `$value` which involved by `annotate_with_types`. Such details should stay on rtk-query enpoints level. All other code should know nothing about `$type`, `$value`.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
ma-efremoff: Looks like a duplicated type from src/shared/yt-types. Is there other duplicates?

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    it('annotates the request body as typed input', async () => {
+        const {body} = bigIntegerKeyBody();
+
+        await flowDeleteStates('//pipeline', body);
+
+        const [, args] = mockFlowExecute.mock.calls[0];
+        expect(args.parameters.input_format).toBe(TYPED_INPUT_FORMAT);
+        expect(args.data).toBe(body);
+    });
+});
```
ma-efremoff: These tests not much useful, it is better to remove them.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.ts
ma-efremoff: Such regexp looks unstable. It is better to have specific fields in response.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.spec.ts
```diff
@@ -0,0 +1,1973 @@
+import {
```
ma-efremoff: It is unfair to split commits just by files. New entities should be in the same commit with their first usage.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateKeyBuilder/FlowStateKeyBuilder.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    column={column}
+                    value={values[column.name] ?? ''}
+                    onChange={(next) => onChange({...values, [column.name]: next})}
+                />
+            ))}
+        </Flex>
+    );
+}
+
+export default FlowStateKeyBuilder;
```
ma-efremoff: We prefer to not use `export default` for components

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateFilters/FlowStateFilters.tsx
```diff
@@ -0,0 +1,246 @@
...
+                            }
+                        />
+                    </Flex>
+                </Card>
+            )}
+        </Flex>
+    );
+}
+
+export default FlowStateFilters;
```
ma-efremoff: !

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.tsx
ma-efremoff: We prefer to not use Table from uikit

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowDeleteStatesDialog/FlowDeleteStatesDialog.tsx
```diff
@@ -0,0 +1,238 @@
...
+    visible,
```
ma-efremoff, line 31: We prefer YTDFDialog for editors/confirmations

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateSection/FlowStateSection.tsx
ma-efremoff: We prefer WithStickyToolbar

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
```diff
@@ -87,6 +133,12 @@ function FlowComputationTabs({computation}: {computation: string}) {
                     url: `/${cluster}/${Page.FLOWS}/computations/${encodeURIComponent(computation)}/monitor`,
                     show: true,
                 },
+                ...extraTabs.map((tab) => ({
+                    value: tab.value,
+                    text: tab.title,
+                    url: `/${cluster}/${Page.FLOWS}/computations/${encodeURIComponent(computation)}/${tab.value}`,
```
ma-efremoff: We prefer to use helpers from src/ui/utils/app-url

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
ma-efremoff: src/ui/utils/app-url

packages/ui/src/ui/pages/flow/Flow/Flow.tsx
ma-efremoff: Why do we need to check it?

blinkov:
Thanks for the thorough review. This is a substantial set of feedback covering the data-fetching pattern (rtk-query vs manual fetch), type handling ($type/$value via ypath vs raw access, duplicate types), component conventions (DataTableGravity, YTDFDialog, WithStickyToolbar, export default), file/component splitting, and test philosophy. Working through it properly now rather than addressing individual comments piecemeal - will follow up with either code changes or clarifying questions on the ones that need more context (e.g. the two `Why do we need the check?` comments).

packages/ui/src/ui/constants/index.ts
blinkov: Done in 272b8af97548a1e8f8cd0034524db917135fe988.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
blinkov: Done in 48ed89c47.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
blinkov: Done in 48ed89c47.

blinkov: Done in 48ed89c47.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.tsx
blinkov: Done in e5e2e6b55.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
blinkov: Done in e5e2e6b55.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
blinkov: Done in e5e2e6b55.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
blinkov: Done in e5e2e6b55.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
blinkov: Done in 9eb978309.

blinkov: Done in 9eb978309.

blinkov: Done in 9eb978309.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/types.ts
blinkov: Done in 9eb978309.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.ts
blinkov: Done in 4f799952a.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.tsx
blinkov: Done in 36c51ad06.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.tsx
blinkov: Done in fcb2e13b8.

blinkov: Done in fcb2e13b8.

blinkov: Done in fcb2e13b8.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateSection/FlowStateSection.tsx
blinkov: Done in e7baa335c.

blinkov: Done in e7baa335c.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateSection/FlowStateSection.tsx
blinkov: Done in 404e33dd7.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateKeyBuilder/FlowStateKeyBuilder.tsx
blinkov: Done in 64880a8e9.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowDeleteStatesDialog/FlowDeleteStatesDialog.tsx
blinkov: Looked at YTDFDialog: it wraps @gravity-ui/dialog-fields' DFDialog, which models a single-submit form (fields + one apply action). This dialog is a two-phase workflow — load pipeline state, gate on it, dry-run a preview, then a commit that is only enabled while the preview still matches the current selection and force flag. DFDialog has no second submit action and no preview-invalidation semantics, so adopting it would mean hiding the state machine inside custom registered controls — more code, less clarity. I kept the uikit Dialog shell and instead split the dialog into a useFlowDeleteStates controller plus RowsSummary / OutcomesSummary / DeleteStatesStatus (c47ded086). If dropping the preview step in favour of a single confirm is acceptable, YTDFDialog fits and I'll switch.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowDeleteStatesDialog/FlowDeleteStatesDialog.tsx
blinkov: Looked at YTDFDialog: it wraps @gravity-ui/dialog-fields' DFDialog, which models a single-submit form (fields + one apply action). This dialog is a two-phase workflow — load pipeline state, gate on it, dry-run a preview, then a commit that is only enabled while the preview still matches the current selection and force flag. DFDialog has no second submit action and no preview-invalidation semantics, so adopting it would mean hiding the state machine inside custom registered controls — more code, less clarity. I kept the uikit Dialog shell and instead split the dialog into a useFlowDeleteStates controller plus RowsSummary / OutcomesSummary / DeleteStatesStatus (c47ded086). If dropping the preview step in favour of a single confirm is acceptable, YTDFDialog fits and I'll switch.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.ts
blinkov: Agreed these were fragile — all four regex literals are gone in 4f799952a, replaced by explicit token parsing in heavy-hitters.ts with tests for the malformed cases. There are no structured fields to switch to, though: heavy hitters arrive as preformatted strings inside FlowMessageType.yson (typed unknown) on describe-computation messages, so Key/Ratio/PartitionId only exist inside that text. Exposing them as structured fields needs a backend change to the heavy-hitters message payload — happy to file that. The parser degrades safely: any line it cannot parse renders as plain text and never produces a wrong link.

blinkov: Agreed these were fragile — all four regex literals are gone in 4f799952a, replaced by explicit token parsing in heavy-hitters.ts with tests for the malformed cases. There are no structured fields to switch to, though: heavy hitters arrive as preformatted strings inside FlowMessageType.yson (typed unknown) on describe-computation messages, so Key/Ratio/PartitionId only exist inside that text. Exposing them as structured fields needs a backend change to the heavy-hitters message payload — happy to file that. The parser degrades safely: any line it cannot parse renders as plain text and never produces a wrong link.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.ts
blinkov: Done in e5e2e6b55: flow-state-api.ts is gone; read-states/delete-states are typed FlowExecuteTypes commands, so both registry casts disappeared; getPipelineState needed no cast once the duplicate local type was replaced with GetPipelineStateData (it is already typed in rum-wrap-api.ts), and getPipelineSpec is typed by the endpoint's declared result. Two narrow assertions remain by design: the `{error} as {error: YTError}` catch (the same idiom as store/api/yt/flow/endpoint.ts) and the single cast where the normalized wire payload is given its response type.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateFilters/FlowStateFilters.tsx
blinkov: This came through as a bare `!` — could you say what change you'd like here? If it flags the redundant `export default` on that line, it's removed in 64880a8e9 along with the other state-viewer default exports.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.spec.ts
blinkov: The commits now ship grouped by logical entity rather than by file, from 48ed89c47 through 64880a8e9: shared types (48ed89c47), rtk-query endpoints (e5e2e6b55), ypath adoption (9eb978309), helpers split (4f799952a), RoutedLink (36c51ad06), DataTableGravity (fcb2e13b8), sticky toolbar (e7baa335c), hooks extraction (404e33dd7), dialog split (c47ded086), named exports (64880a8e9). Each commit is one self-contained change across whatever files it touches.

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
blinkov: Removed in 7f082827a — confirmed with the author that it was an unverified precaution.

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
blinkov: Removed in 7f082827a — confirmed with the author that it was an unverified precaution.

packages/ui/src/ui/pages/flow/Flow/Flow.tsx
blinkov: Removed in 7f082827a — confirmed with the author that it was an unverified precaution.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowStateFilters/FlowStateFilters.tsx
blinkov: Checked the exact anchor of this comment (original_line 246 of the file as first added in 48a8e7d01) - it lands precisely on `export default FlowStateFilters;`, the last line of the file. Same complaint as the export-default comment on FlowStateKeyBuilder.tsx - confirmed and removed in 64880a8e9 along with the other state-viewer default exports.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.spec.tsx
blinkov: Went through every remaining spec case and split them by what the assertion actually is.

Converted to the storybook + screenshot pattern in fb8980e95: `FlowStateResults` now has `__stories__/FlowStateResults.stories.tsx` (Default / NoResults / WithResponseErrors) and `__tests__/FlowStateResults.visual.test.tsx` with committed snapshots, matching the existing `NoContent` / dashboard-widget idiom. That replaces `FlowStateResults.rowLink.spec.tsx`, which was deleted — its two cases only asserted computed CSS (`display: inline-flex`, `align-items: center`) on hand-rewritten markup, and the screenshot of a hovered row now pins that alignment on the real component instead.

To make this possible the playwright mount fixture needed `window.store` to be set (`RoutedLink` resolves hrefs through `makeRoutedURL`, which reads it); without it any component rendering a `RoutedLink` mounted blank. That is a two-line change in `playwright-components/playwright/index.tsx` plus a narrow `setWindowStore()` export, so page-level components can now be covered this way in general.

The rest I kept as jest, because a screenshot cannot express what they assert:
- `FlowComputationMessages.spec.tsx` — the generated hrefs and the JSON `heavyHitterSeed` query param, and that the heavy-hitters message is filtered out of the raw list.
- `FlowDeleteStatesDialog.spec.tsx` — the preview-before-delete gate, the force-toggle invalidating a preview, and the exact mutation body/callbacks.
- `FlowStateFilters.spec.tsx` — that clicking a disabled option does not fire `onChange`, and the reconciled value passed to `onChange`; plus `aria-disabled` / `title` semantics, which are accessibility attributes rather than pixels.

All three also depend on rtk-query hooks whose deterministic story setup would need msw handlers, and the payoff there would still only be a greyed-out option or a disabled button, not the behaviour under test.

Verified locally: `npm run test:components` green (8 passed, chromium + webkit, light + dark, via `scripts/docker-run.sh`), `npm run test:unit` 701 passed / 28 suites, `npm run lint` clean, `npm run typecheck` unchanged from the pre-existing 7 `@ytsaurus/components-typings` errors in 6 unrelated MetaTable/query-tracker files.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/flow-state-api.spec.ts
blinkov: This file has since moved to `packages/ui/src/ui/store/api/yt/flow/flow-state.spec.ts` (78 lines). I kept it as jest rather than removing it.

It is not a component test — it asserts the rtk-query endpoint contract: the request bodies built for `read_states` / `delete_states`, and how the responses are normalised. There is no rendering involved, so there is nothing a screenshot could cover, and dropping it would leave the request-shape wiring untested.

The same applies to `FlowStateViewer/helpers.spec.ts`, which is pure parsing and state-shape logic.

Where the assertions genuinely were about rendering I did switch to the pattern you asked for — see the reply on the `FlowComputationMessages` thread: `FlowStateResults` got stories plus a `.visual.test.tsx` with committed screenshots in fb8980e95, and `FlowStateResults.rowLink.spec.tsx` was deleted as a result.

Happy to remove these two as well if you would rather not have endpoint-contract coverage in the PR at all — just say the word and I will drop them.

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
blinkov: Done in f3c12583347f80ccaa9184e737cab347f37d7ddc — added `makeFlowComputationTabLink` in `src/ui/utils/app-url/makeFlowLink.ts` and switched `FlowComputationTabs` to build all three tab URLs (details/monitor/extra tabs) through it instead of hand-rolled template strings.

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
blinkov: Done in f3c12583347f80ccaa9184e737cab347f37d7ddc — same fix, both this line and the sibling comment on line 129 are addressed by the new `makeFlowComputationTabLink` helper.

packages/ui/src/ui/pages/flow/Flow/FlowComputation/FlowComputation.tsx
blinkov: Done in f3c12583347f80ccaa9184e737cab347f37d7ddc (duplicate of the other thread on this same line — same fix applies).

packages/ui/src/ui/pages/flow/Flow/Flow.tsx
blinkov: Removed in 7f082827a — the isLocalMode gate was verified not load-bearing (the state tab now shows unconditionally) and dropped entirely, so this check no longer exists.

packages/ui/babel.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
             presets: [
                 ['@babel/preset-env', {targets: {node: 'current'}}],
                 '@babel/preset-typescript',
+                ['@babel/preset-react', {runtime: 'automatic'}],
```
ma-efremoff: Why do we need it?

packages/ui/package.json
```diff
@@ -101,6 +101,7 @@
     "zod": "^4.4.3"
   },
```
ma-efremoff, line 102: All the changes in the file looks unnecessary.

packages/ui/src/ui/store/window-store.ts
```diff
@@ -1,13 +1,18 @@
 import {type AppBrowserHistory, type StoreType} from './store.main';
 import {type Store} from 'redux';
 
-export function setWindowStoreAndHistory(store: Store, appBrowserHistory: AppBrowserHistory) {
-    Object.assign(window, {store, appBrowserHistory});
+export function setWindowStore(store: Store) {
```
ma-efremoff: Is the function used anywhere?

packages/ui/src/ui/playwright-components/playwright/index.tsx
```diff
@@ -37,6 +38,7 @@ beforeMount(async ({App}) => {
             getDefaultMiddleware({serializableCheck: false}).concat(rootApi.middleware),
         devTools: process.env.NODE_ENV !== 'production',
     });
+    setWindowStore(store);
```
ma-efremoff: Why do we need it here?

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.spec.ts
```diff
@@ -0,0 +1,2093 @@
...
```
ma-efremoff: There is no helpers.ts to have helpers.spec.ts now

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/state-filters.ts
```diff
@@ -0,0 +1,491 @@
...
+function decodeRawKeyToken(token: string, column: FlowKeyColumn): string | undefined {
+    const trimmed = token.trim();
+    if (!trimmed) {
+        return undefined;
+    }
+    if (trimmed.includes('"')) {
+        return column.type === 'string' ? decodeQuotedRawKeyToken(trimmed) : undefined;
+    }
+    if (column.type === 'string') {
+        return /^[A-Za-z_][A-Za-z0-9_.-]*$/.test(trimmed) ? trimmed : undefined;
```
ma-efremoff: Why do we need such restriction for strings?

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/state-values.ts
```diff
@@ -0,0 +1,269 @@
...
+function replaceAnnotatedIntegers(_key: string, value: unknown): unknown {
+    return isAnnotatedBigInteger(value) ? value.$value : value;
+}
+
+export function decodeStateString(value: string): string {
+    if ([...value].some((character) => character.charCodeAt(0) > 0xff)) {
+        return value;
+    }
+    try {
+        return unipika.utils.utf8.decode(value);
```
ma-efremoff: Can we move such decoding on request-level?
Additionally we have JSONSerializer for javascript-wrapper might it be useful for the PR?

packages/ui/babel.config.js
blinkov: Agreed. This preset was introduced only for the added Jest TSX component tests. I propose removing the added `.spec.tsx` suites or moving their rendering coverage to Playwright component stories, then dropping `@babel/preset-react` and this Babel change.

packages/ui/package.json
blinkov: Agreed. These dependencies support the added Jest TSX component tests. I propose removing the TSX Jest suites or migrating rendering assertions to Playwright component tests, retaining only non-rendering `.spec.ts` coverage, and then removing the added React/jsdom dependencies and lockfile changes.

packages/ui/src/ui/store/window-store.ts
blinkov: It is currently used once by the Playwright component fixture to initialize `window.store` for components containing `RoutedLink`. I propose removing the standalone helper if that fixture can use the normal application initializer; otherwise I will keep the behavior but avoid widening the production store API solely for tests.

packages/ui/src/ui/playwright-components/playwright/index.tsx
blinkov: It was added because `RoutedLink` calls URL helpers that read `window.store`, and affected component stories otherwise mount blank. I propose switching the fixture to the normal store initialization path (or a test-local setup) so Playwright support does not require a production API change.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.spec.ts
blinkov: Agreed. I propose splitting this suite by its current implementation modules (`state-filters`, `state-requests`, `state-delete`, `state-values`, and `heavy-hitters`) and keeping only focused non-rendering `.spec.ts` tests beside each module.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/state-filters.ts
blinkov: The restriction was intended to prevent an unquoted string from being mistaken for YSON syntax. I propose replacing this handwritten bare-string restriction with the project YSON parser/serializer behavior, so valid escaped or quoted UTF-8 keys round-trip without an artificial identifier-only grammar.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/state-values.ts
blinkov: Agreed on the layer boundary. I propose moving recursive key-string decoding into the read-states endpoint normalization, evaluating the existing javascript-wrapper JSONSerializer there, and leaving UI helpers to consume already-normalized values. I will preserve annotated 64-bit integers exactly and cover UTF-8 key click/copy/filter round-trips.

packages/ui/babel.config.js
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676.

packages/ui/package.json
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676.

packages/ui/src/ui/store/window-store.ts
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676.

packages/ui/src/ui/playwright-components/playwright/index.tsx
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.spec.ts
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/state-filters.ts
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/state-values.ts
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.spec.tsx
blinkov: Updated in 3efdba7c9523558a714c8b2c5549729dedcca676: all rendered `*.spec.tsx` suites are removed; rendering coverage now uses Playwright component stories, while non-rendering behavior is covered by focused pure `*.spec.ts` suites.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowDeleteStatesDialog/FlowDeleteStatesDialog.tsx
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676: deletion now uses `YTDFDialog`, with the safety gates, acknowledgement, partial outcomes, and retry behavior preserved.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowDeleteStatesDialog/FlowDeleteStatesDialog.tsx
blinkov: Done in 3efdba7c9523558a714c8b2c5549729dedcca676: deletion now uses `YTDFDialog`, with the safety gates, acknowledgement, partial outcomes, and retry behavior preserved.

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.spec.ts
blinkov: Updated in 3efdba7c9523558a714c8b2c5549729dedcca676: the branch history is rebuilt as one cohesive feature commit, and the obsolete aggregate `helpers.spec.ts` is split into focused suites beside each implementation.

packages/ui/babel.config.js
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/package.json
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/store/window-store.ts
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/playwright-components/playwright/index.tsx
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.spec.ts
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/state-filters.ts
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/state-values.ts
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowComputationMessages/FlowComputationMessages.spec.tsx
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowDeleteStatesDialog/FlowDeleteStatesDialog.tsx
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/FlowDeleteStatesDialog/FlowDeleteStatesDialog.tsx
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

packages/ui/src/ui/pages/flow/flow-components/FlowStateViewer/helpers.spec.ts
blinkov: Final amended commit: bb33746cc1704dba4d20a4788924b985c476c411 (supersedes 3efdba7c9523558a714c8b2c5549729dedcca676).

PR #1754, opened 2026-08-14
feat(Flow/Graph): add partitions distribution dialog for computation nodes [#1753]

### LINKED ISSUE #1753: Flow: no way to visually check partition distribution uniformity on the Graph page
The Flow pipeline UI exposes per-partition metrics (cpu_usage, memory_usage, messages_per_second, bytes_per_second) only as tables: the per-computation partitions list and the Total/Avg/Min/Max summary. There is no visualization of the distribution over partitions, so checking that partitions are loaded uniformly (no skewed/hot partitions) requires sorting the table by hand.

Proposal: add a chart-icon button next to the partition counter on each computation node of the **Graph** page that opens a dialog with a histogram of the selected metric over all partitions of that computation, with a metric switcher and the computation average drawn as a plot line. Data comes from the existing `describe-computation` command, so no backend changes are needed.

A PR implementing this is coming shortly.

timoninmaxim:
https://nda.ya.ru/t/mzLolgBR7n5kKi
 Closes #1753.

Adds a chart-icon button next to the partition counter on every computation node of the Flow **Graph** page (both the node overlay and the hover popup). Clicking it opens a dialog with a histogram of the selected metric over all partitions of that computation:

- metric switcher: CPU usage / Memory usage / Messages per second / Bytes per second;
- the computation average is drawn as a vertical plot line;
- degenerate cases (no partitions; all values equal / single partition) show a text summary instead of a chart.

Data comes from the existing `describe-computation` command — no backend changes. The dialog follows the `FlowMessagesDialogContext` pattern (provider at `FlowPage` root), the chart reuses `YTChartKitHistogram`. The button is rendered only for real computation nodes, not for computation groups.

Based on `ui-v3.21.1` (the current arc `ytsaurus_base_branch` pin) so an internal stand can build against this PR head.

## Summary by Sourcery

Add a partitions distribution dialog for individual computations on the Flow Graph, showing per-partition metrics in a histogram with a metric selector and average line, opened via a chart button on computation nodes.

New Features:
- Introduce a Flow partitions distribution dialog that visualizes per-partition metrics (CPU, memory, messages per second, bytes per second) for a selected computation.
- Add a chart icon control to computation nodes in the Flow Graph to open the partitions distribution dialog when partition stats are available.

Enhancements:
- Wire a dedicated FlowPartitionsDistributionDialogContext at the Flow page level to manage visibility and data fetching for the new distribution dialog.
- Reuse existing describe-computation query and YTChartKitHistogram component to display partition distributions, with graceful handling of empty or degenerate datasets.

### DIFF
--- a/packages/ui/src/ui/pages/flow/Flow/FlowGraph/FlowGraph.tsx
+++ b/packages/ui/src/ui/pages/flow/Flow/FlowGraph/FlowGraph.tsx
@@ -198,7 +198,14 @@ function FlowGraphDataButton() {
 function renderContent({item, ...rest}: {item: FlowGraphBlock; detailed?: boolean}) {
     switch (item.is) {
         case 'computation':
-            return <Computation className={block('item')} item={item} {...rest} />;
+            return (
+                <Computation
+                    className={block('item')}
+                    item={item}
+                    showPartitionsDistribution
+                    {...rest}
+                />
+            );
         case 'stream':
             return <Stream className={block('item')} item={item} {...rest} />;
         case 'computation-group':
--- a/packages/ui/src/ui/pages/flow/Flow/FlowGraph/i18n/en.json
+++ b/packages/ui/src/ui/pages/flow/Flow/FlowGraph/i18n/en.json
@@ -4,5 +4,6 @@
   "context_high-consumption": "High consumption",
   "field_zoom-to": "Zoom to:",
   "action_graph-data": "Graph data",
-  "action_messages": "Messages ({{count}})"
+  "action_messages": "Messages ({{count}})",
+  "action_partitions-distribution": "Partitions distribution"
 }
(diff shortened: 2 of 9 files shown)


timoninmaxim:
Addressed the review:

1. **Gate the icon on having partitions** — applied: the trigger now renders only when `partitions_stats.count > 0`.
2. **Accessibility** — applied, but not with the literal suggestion: `ClickableText` forwards only `className`/`color`/`onClick`/`children` to the DOM, so `aria-label`/`title` on it would silently do nothing. Replaced the trigger with a flat `Button` (`view="flat-secondary"`, `size="xs"`) carrying `extraProps={{'aria-label': …}}`, which is the pattern already used elsewhere in the repo.
3. **Renaming `setVisibleDistribution`** — keeping as is: the name deliberately mirrors the sibling `FlowMessagesDialogContext.setVisibleMessages` pattern in the same directory, and consistency between the two dialog contexts seems more valuable here than a more literal name.

timoninmaxim:
Superseded by #1757 — identical commit, re-opened from a branch in this repo so the workflows run without approval and the internal stand build can resolve the branch. The Sourcery review feedback was already addressed here (see the reply above) and is included in #1757.

PR #1757, opened 2026-08-18
feat(Flow): add partitions distribution for computations [#1753]

### LINKED ISSUE #1753: Flow: no way to visually check partition distribution uniformity on the Graph page

timoninmaxim:
https://nda.ya.ru/t/mzLolgBR7n5kKi
         

Closes #1753. Supersedes #1754.

Shows how the load of a computation is spread over its partitions, to spot data skew. Entry points: a chart button next to the partition counter on Flow Graph nodes (computation groups included), the same button in the computations table, and a **Partitions distribution** tab on the computation page.

- Metrics: CPU usage, memory usage, messages per second, bytes per second, each with a short note on how it is measured.
- Views: a bar per partition sorted by value (default), a histogram with a fixed bar width so charts of one metric are comparable, and heavy hitters (keys taking a noticeable share of their partition's input).
- Window for CPU and memory: now / 30 s / 10 min. The 10 min values come from `describe-computation`, the others from `get-flow-view` (`/feedback/partition_job_statuses`). No backend changes.
- min / average / max in the caption and the heavy-hitter rows link to the partition in the computations table.

`YTChartKitHistogram` gets optional props (`barWidth`, `seriesName`, `yAxisMin`, axis titles); without them it renders as before, except zoom is enabled. `YTChartKitBars` is new, for the per-partition view. The `/distribution` route is registered in `location.main.ts`.


packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
```diff
@@ -36,67 +65,77 @@ export function YTChartKitHistogram(props: YTChartKitHistogramProps) {
...
                 // TODO: uncomment the line below when https://github.com/gravity-ui/charts/issues/87 is ready
-                // max,
+                // max: to,
             },
             yAxis: [
                 {
-                    min,
-                    // TODO: uncomment the line below when https://github.com/gravity-ui/charts/issues/87 is ready
-                    // max: plotLinesMax,
+                    min: 0,
```
ma-efremoff: We don't need such hardcoded value. It is better to allow forward specific value throug props, like:
```ts
min: props.min ?? min,
```

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
```diff
@@ -36,67 +65,77 @@ export function YTChartKitHistogram(props: YTChartKitHistogramProps) {
...
+            step = barWidth * Math.max(1, times);
+            count = Math.max(1, Math.ceil((effectiveMax - from) / step));
+        }
 
-        const step = (effectiveMax - min) / barCount;
         let maxSum = 0;
         const values = data.reduce(
             (acc, v) => {
-                const index = Math.floor((v - min) / step);
+                const index = Math.min(count - 1, Math.floor((v - from) / step));
```
ma-efremoff: Please provide some comments why do we need the change

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
```diff
@@ -120,26 +159,24 @@ export function YTChartKitHistogram(props: YTChartKitHistogramProps) {
...
                     return (
                         <React.Fragment>
                             <div>
                                 <ColorCircle color={color ?? 'magenta'} />
-                                {i18n('context_observations-in-range', {
-                                    count: <b key="count">{format.Number(y, {digits: 0})}</b>,
-                                    from: <b key="from">{fmt(l)}</b>,
-                                    to: <b key="to">{fmt(r)}</b>,
-                                })}
+                                {name}: <b>{format.Number(y, {digits: 0})}</b>
```
ma-efremoff: As I understand you need to provide custom tooltip formatter. It is better to keep `i18n(`context_observations-in-range`, ...)` call here for backward compatibility. And add corresponding property for your needs.

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
```diff
@@ -18,13 +18,42 @@ export type YTChartKitHistogramProps = {
     /** by default it uses 10 */
     barCount?: number;
     xPlotLines?: Record<string, number | undefined>;
+    /**
+     * Fixed width of a bar, the bars are aligned to zero.
+     * When the values do not fit into |maxBarCount| bars the width grows by a whole number of times.
+     */
+    barWidth?: number;
```
ma-efremoff, line 24: Isn't it enough to use barCount?

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
timoninmaxim: Done: added `yAxisMin?: number`, defaulting to the previous data minimum; the flow charts pass `yAxisMin={0}`.

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
timoninmaxim: Done, added comments: with `barWidth` the bars have a fixed width and start at zero, and the last bar is half-open, so the clamp keeps a value rounded past `effectiveMax` inside it.

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
timoninmaxim: Done: `context_observations-in-range` is back as the default tooltip, the two-row variant is used only with `seriesName`. Without the new props the component renders as before; the only change for existing callers is the enabled zoom.

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
timoninmaxim: Agreed, dropped `maxBarCount`. `barCount` is the exact number of bars by default and the upper bound with `barWidth` (the width grows by a whole number of times until the bars fit); flow passes `barCount={50}`.

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
```diff
@@ -15,17 +15,15 @@ import i18n from './i18n';
 
 export type YTChartKitHistogramProps = {
     data?: Array<number>;
-    /** by default it uses 10 */
+    /** by default it uses 10, with |barWidth| it is the maximum number of bars */
     barCount?: number;
```
ma-efremoff: It is better to explicitly name it `maxBarCount`, then we don't need such details in comments. And it is better to put the property as close as possible to `barWidth`.

packages/ui/src/ui/pages/flow/Flow/FlowGraph/renderers/FlowPartitionsDistributionDialogContext/FlowPartitionsDistributionDialogContext.tsx
```diff
@@ -301,6 +346,91 @@ function PartitionsCharts({items, metric, view}: PartitionsProps & {items: Array
...
+    children,
+}: {
+    computationId: string;
+    partitionId: string;
+    children: React.ReactNode;
+}) {
+    const path = useSelector(selectFlowPipelinePath);
+    const {setVisibleDistribution} = useFlowPartitionsDistributionDialogContext();
+    return (
+        <Link
```
ma-efremoff: Link component is marked as depricated, it is better to take RoutedLink

packages/ui/src/ui/components/YTChartKit/YTChartKitHistogram.tsx
timoninmaxim: Done: renamed `barCount` to `maxBarCount` and moved it next to `barWidth`.

packages/ui/src/ui/pages/flow/Flow/FlowGraph/renderers/FlowPartitionsDistributionDialogContext/FlowPartitionsDistributionDialogContext.tsx
timoninmaxim: Done, switched to `RoutedLink`.

ma-efremoff (approved):

PR #1760, opened 2026-08-20
feat(Scheduling): allow custom pool metadata rendering [YTFRONT-5882]

ma-efremoff:
https://nda.ya.ru/t/w_D1Lpa87nnuua
            ## Summary by Sourcery

Enable customizable scheduling pool metadata rendering and standardize cluster-aware table-reading requests.

New Features:
- Add an extension point for custom scheduling pool metadata table rendering while retaining the default metadata table.
- Add a reusable read-table API query with cluster-aware request handling and polling support.

Bug Fixes:
- Preserve interactive workload metadata for static table reads while removing it from unrelated table-loading paths.
- Improve cluster and proxy propagation across YT API queries and mutations.

Enhancements:
- Add finite-number type validation and support workload categories in read-table parameters.
- Add not-found error classification for YT API responses.


packages/ui/src/ui/pages/scheduling/Content/tabs/Overview/SchedulingPoolMetaTable/SchedulingPoolMetaTable.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+import {MetaTable, type MetaTableProps} from '@ytsaurus/components';
+
+import UIFactory from '../../../../../../UIFactory';
+
+export type SchedulingPoolMetaTableProps = MetaTableProps;
+
+export function SchedulingPoolMetaTable(props: SchedulingPoolMetaTableProps) {
+    return UIFactory.renderSchedulingPoolMetaTable(props) || <MetaTable {...props} />;
```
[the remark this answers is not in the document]
SimbiozizV: This isn’t a blocker, but `renderSchedulingPoolMetaTable` can indeed return a valid falsy React node, which will be replaced with `MetaTable`.

packages/ui/src/ui/store/api/yt/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function getEffectiveClusterArgs<T extends YTEndpointApiArgs<unknown>>(
+    args: T,
+    currentCluster: string,
+): T {
+    const {cluster, setup, ...rest} = args;
+
+    const effectiveCluster = cluster ?? currentCluster;
+    const effectiveSetup = effectiveCluster
+        ? {proxy: getClusterProxy(getClusterConfigByName(effectiveCluster)), ...setup}
```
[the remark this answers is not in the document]
SimbiozizV: There is indeed quite a cunning, non‑obvious logic at play here.

Right now, the condition is tied to effectiveCluster, which isn’t entirely accurate. The check for setup.proxy is important because, if it’s present, the result of the getClusterProxy calculation will be replaced regardless of the result obtained.

packages/ui/src/ui/store/api/yt/types.ts
```diff
@@ -12,12 +12,14 @@ import {type ApiMethodParams, type YTApiSetup} from '../../../rum/rum-wrap-api';
 
 // YT API Utility Types
 
-export type YTEndpointApiArgs<CommandParameters> = ClusterOrSetup &
-    Omit<ApiMethodParams<CommandParameters>, 'setup'>;
+export type YTEndpointApiArgs<CommandParameters> = ApiMethodParams<CommandParameters> & {
+    cluster?: string;
+};
 
 export type ClusterOrSetup =
```
SimbiozizV: It seems that after the change to `YTEndpointApiArgs`, this type is no longer used anywhere.

packages/ui/src/ui/store/actions/navigation/content/table/readStaticTable.ts
```diff
@@ -13,15 +14,22 @@ import {
...
-}: ReadTableParameters<{path: string}>): Promise<ReadTableResult> {
-    const tmp = await ytApiV3.readTable({
+}: ReadStaticTableParameters): Promise<ReadTableResult> {
+    const tmp = await ytApiV3Id.readTable(id, {
         setup: {...setup, ...tableReadSetup},
-        parameters: {...parameters, ...tableReadParameters},
+        parameters: {
+            ...parameters,
+            ...tableReadParameters,
+            ...{table_reader: {workload_descriptor: {category: 'user_interactive'}}},
```
SimbiozizV: Here, you don’t need `...`. You can use `table_reader: {workload_descriptor: {category: 'user_interactive'}}`.

packages/ui/src/ui/store/api/yt/types.ts
```diff
@@ -12,12 +12,14 @@ import {type ApiMethodParams, type YTApiSetup} from '../../../rum/rum-wrap-api';
...
     // cluster or setup param should be required for the case of selection
-    {cluster?: string} | {setup: Omit<YTApiSetup, 'proxy'> & Pick<Required<YTApiSetup>, 'proxy'>};
+    | {cluster?: string; setup?: never}
+    | {setup: Omit<YTApiSetup, 'proxy'> & Pick<Required<YTApiSetup>, 'proxy'>; cluster?: never};
```
xexys: I think it’s better to list the fields in the same order.

xexys: Is it possible to simplify the type to `Override<YTApiSetup, {proxy: string}>` ?

packages/ui/src/ui/utils/errors/index.ts
```diff
@@ -38,6 +38,10 @@ export function getPermissionDeniedError<T extends YTErrorRaw>(error: T): T | un
     return getErrorWithCode([error], yt.codes.PERMISSION_DENIED);
 }
 
+export function getNotFoundError<T extends YTErrorRaw>(error: T): T | undefined {
+    return getErrorWithCode([error], yt.codes.NOT_FOUND);
+}
```
xexys, line 43: It seems that using a generic type in this case is actually the wrong approach, because it describes not the error with the code we’re looking for, but the original argument.

packages/ui/src/ui/store/api/yt/readTable/index.ts
```diff
@@ -0,0 +1,36 @@
...
+        readTable: build.query({
+            queryFn: readTable,
+            providesTags: (_result, _error, args) => [args.id],
+        }),
+    }),
+});
+
+type ReadTableResponse<RowT extends Record<string, unknown>> = Omit<ReadTableResult, 'rows'> & {
+    rows: Array<RowT>;
+};
```
xexys: You can use `import {type Override} from 'utility-types';`.
The code becomes more readable.

xexys (changes requested):

packages/ui/src/ui/store/api/yt/utils.ts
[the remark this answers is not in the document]
ma-efremoff: Explicit setup.proxy has upper priority

packages/ui/src/ui/utils/errors/index.ts
ma-efremoff: T is used as part of returning type

xexys (approved):

PR #1762, opened 2026-08-21
fix(Queries): double scroll [YTFRONT-5943]

SimbiozizV:
https://nda.ya.ru/t/eAVPsP4g7nsCAn
 ## Summary by Sourcery

Integrate query history pagination with the result lists to provide a single, consistent infinite-scroll experience.

New Features:
- Support loading additional query history directly as users scroll through both the history table and full-text search results.

Bug Fixes:
- Prevent duplicate or competing pagination controls from causing double scrolling in query history views.

Enhancements:
- Move query list column selection logic into a dedicated UI selector and preserve required columns across presets.

### DIFF
--- a/packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/Columns/selectQueryListColumns.ts
+++ b/packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/Columns/selectQueryListColumns.ts
@@ -1,0 +1,32 @@
+import {createSelector} from 'reselect';
+import intersectionBy_ from 'lodash/intersectionBy';
+import {QueriesListAuthorFilter} from '../../../../../types/query-tracker/queryList';
+import {
+    selectQueriesFilters,
+    selectQueryListHistoryColumns,
+} from '../../../../../store/selectors/query-tracker/queriesList';
+import {ActionColumns, AllColumns, AuthorColumns, MyColumns, NameColumns} from './columns';
+
+const ALL_COLUMN_NAMES = intersectionBy_(AllColumns, MyColumns, 'name').map((item) => item.name);
+const EXCLUDED_COLUMNS = [NameColumns.name, AuthorColumns.name, ActionColumns.name];
+
+export const selectQueryListColumns = createSelector(
+    [selectQueriesFilters, selectQueryListHistoryColumns],
+    (filter, selectedColumns) => {
+        const currentColumnsPreset =
+            filter.user === QueriesListAuthorFilter.My ? MyColumns : AllColumns;
+
+        const selectedColumnNames = new Set(selectedColumns ?? ALL_COLUMN_NAMES);
+
+        selectedColumnNames.add(NameColumns.name);
+        selectedColumnNames.add(AuthorColumns.name);
+        selectedColumnNames.add(ActionColumns.name);
+
+        return {
+            columns: currentColumnsPreset.filter(({name}) => selectedColumnNames.has(name)),
+            allowedColumns: currentColumnsPreset
+                .filter((item) => !EXCLUDED_COLUMNS.includes(item.name))
+                .map(({name}) => ({name, checked: selectedColumnNames.has(name)})),
+        };
+    },
+);
--- a/packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/FullTextSearch/FullTextSearch.tsx
+++ b/packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/FullTextSearch/FullTextSearch.tsx
@@ -1,7 +1,8 @@
-import React, {type FC} from 'react';
+import React, {type FC, useCallback} from 'react';
 import {Flex, List, Loader} from '@gravity-ui/uikit';
-import {useSelector} from '../../../../../store/redux-hooks';
+import {useDispatch, useSelector} from '../../../../../store/redux-hooks';
 import {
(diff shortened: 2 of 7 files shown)


packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/HistoryList/HistoryList.tsx
```diff
@@ -44,7 +53,8 @@ export function HistoryList() {
                 itemHeight={LIST_ITEM_HEIGHT}
                 itemsHeight={LIST_ITEM_HEIGHT * items.length}
                 items={items}
-                loading={isLoading && items.length === 0}
+                loading={(isLoading && items.length === 0) || showPagination}
+                onLoadMore={showPagination ? handleLoadMore : undefined}
```
xexys: I don't understand the current logic a bit. 

How can the `showPagination` feature determine whether the `HistoryListHeader` list is in the loading state? 

How can the `showPagination' feature affect the loading of new items via `onLoadMore'? It seems that in this case `showPagination` should be referred to as `hasNextPage`, but it also does not work with the case above.

xexys: It is possible that there is a problem with the names of variables, and perhaps we need to find something more relevant.

xexys (changes requested):

packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/FullTextSearch/FullTextSearch.tsx
```diff
@@ -48,6 +57,8 @@ export const FullTextSearch: FC = () => {
             itemsHeight={items.length * LIST_ITEM_HEIGHT}
             filterable={false}
             items={prepareFullTextSearchItems({items, filter, maxLines: MAX_PREVIEW_LINES})}
+            loading={hasNextPage}
```
xexys, line 60: Maybe we should use `isLoading` here?

packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/HistoryList/HistoryList.tsx
```diff
@@ -44,7 +54,8 @@ export function HistoryList() {
                 itemHeight={LIST_ITEM_HEIGHT}
                 itemsHeight={LIST_ITEM_HEIGHT * items.length}
                 items={items}
-                loading={isLoading && items.length === 0}
+                loading={isInitialLoading || hasNextPage}
```
xexys, line 57: Why does `hasNextPage` affect whether the list is being loaded at the moment or not?

packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/Columns/selectQueryListColumns.ts
```diff
@@ -0,0 +1,35 @@
...
+} from '../../../../../store/selectors/query-tracker/queriesList';
+import {ActionColumns, AllColumns, AuthorColumns, MyColumns, NameColumns} from './columns';
+
+export const selectQueryListColumns = createSelector(
+    [selectQueriesFilters, selectQueryListHistoryColumns],
+    (filter, selectedColumns) => {
+        const ALL_COLUMN_NAMES = intersectionBy_(AllColumns, MyColumns, 'name').map(
+            (item) => item.name,
+        );
+        const EXCLUDED_COLUMNS = [NameColumns.name, AuthorColumns.name, ActionColumns.name];
```
xexys: Is it possible to move these constants outside the selector so that they don’t have to be calculated every time?

packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/Columns/selectQueryListColumns.ts
```diff
@@ -0,0 +1,35 @@
...
+        const currentColumnsPreset =
+            filter.user === QueriesListAuthorFilter.My ? MyColumns : AllColumns;
+
+        const selectedColumnNames = new Set(
+            Array.isArray(selectedColumns) ? selectedColumns : ALL_COLUMN_NAMES,
```
xexys: I believe the type here is `Array | undefined`?
It looks like we need to use this condition: `selectedColumns ? selectedColumns : ALL_COLUMN_NAMES`

packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/FullTextSearch/FullTextSearch.tsx
SimbiozizV: If you set `isLoading`, after the first page, sentinel will disappear and the infinite scroll will stop. We want to keep scrolling until we’ve shown everything.

packages/ui/src/ui/pages/query-tracker/QueriesList/QueriesHistoryList/HistoryList/HistoryList.tsx
SimbiozizV: `isInitialLoading` — an empty list on the first request.
`hasNextPage` — sentinel for the next page.

xexys (approved):

PR #1763, opened 2026-08-21
fix(Operations): progress can be null [YTFRONT-5928]

SimbiozizV:
https://nda.ya.ru/t/B_5-_D_x7nsQAv
   ## Summary by Sourcery

Prevent operation views from failing when progress data is null by treating unavailable progress fields as undefined.

Bug Fixes:
- Handle null operation progress safely across operation details, selectors, statistics, job utilities, data-flow views, and task views.

Enhancements:
- Make task and statistics preparation resilient when progress data or nested counters are unavailable.

### DIFF
--- a/packages/ui/src/ui/pages/operations/OperationDetail/tabs/job-sizes/JobSizes/JobSizes.tsx
+++ b/packages/ui/src/ui/pages/operations/OperationDetail/tabs/job-sizes/JobSizes/JobSizes.tsx
@@ -21,6 +21,7 @@ import {selectOperationDetailsLoadingStatus} from '../../../../../../store/selec
 import {useAppRumMeasureStart} from '../../../../../../rum/rum-app-measures';
 import {RumMeasureTypes} from '../../../../../../rum/rum-measure-types';
 import {isFinalLoadingStatus} from '../../../../../../utils/utils';
+import {isNull} from '../../../../../../utils';
 import {useRumMeasureStop} from '../../../../../../rum/RumUiContext';
 import {type RootState} from '../../../../../../store/reducers';
 import YTHistogram, {
@@ -170,7 +171,8 @@ class JobSizes extends React.Component<Props, State> {
 
     renderContent() {
         const {operation} = this.props;
-        const tasks = ypath.getValue(this.props.operation, '/@progress/tasks');
+        const progress = ypath.getValue(this.props.operation, '/@progress');
+        const tasks = isNull(progress) ? undefined : ypath.getValue(progress, '/tasks');
         return (
             <ErrorBoundary>
                 <div className={block()}>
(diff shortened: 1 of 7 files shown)


packages/ui/src/ui/pages/operations/OperationDetail/tabs/job-sizes/JobSizes/JobSizes.tsx
```diff
@@ -170,7 +170,8 @@ class JobSizes extends React.Component<Props, State> {
 
     renderContent() {
         const {operation} = this.props;
-        const tasks = ypath.getValue(this.props.operation, '/@progress/tasks');
+        const progress = ypath.getValue(this.props.operation, '/@progress') || undefined;
+        const tasks = ypath.getValue(progress, '/tasks');
```
xexys: If I understood correctly, the `getValue` function cannot read the full path, but it can read the path from the `undefined` value.

I would suggest writing the code in a different way. First, check that there is a `/@progress`, and if there is one, then read `/@progress/tasks`.

In the current implementation, the code looks more like some kind of magic that takes into account the internal implementation of `getValue`.

Maybe `getValue` shouldn't fall at all, but should just return `undefined`, which is actually what is happening now when splitting into two calls.

packages/ui/src/ui/pages/operations/selectors.ts
```diff
@@ -177,7 +177,7 @@ export class ListOperationSelector extends OperationSelector {
         this.finishTime = ypath.getValue(attributes, '/finish_time');
         this.duration = (moment(this.finishTime) as any) - (moment(this.startTime) as any);
 
-        const progress = ypath.getValue(attributes, '/brief_progress');
+        const progress = ypath.getValue(attributes, '/brief_progress') || undefined;
```
xexys: Why is it explicitly necessary to result in `undefined`?

xexys (changes requested):

packages/ui/src/ui/pages/operations/OperationDetail/tabs/job-sizes/JobSizes/JobSizes.tsx
```diff
@@ -21,6 +21,7 @@ import {selectOperationDetailsLoadingStatus} from '../../../../../../store/selec
 import {useAppRumMeasureStart} from '../../../../../../rum/rum-app-measures';
 import {RumMeasureTypes} from '../../../../../../rum/rum-measure-types';
 import {isFinalLoadingStatus} from '../../../../../../utils/utils';
+import {isNull} from '../../../../../../utils/index';
```
xexys: Maybe `import {isNull} from '../../../../../../utils';` ?

packages/ui/src/ui/pages/operations/selectors.ts
```diff
@@ -4,6 +4,7 @@ import moment from 'moment';
 
 import ypath from '../../common/thor/ypath';
 import hammer from '../../common/hammer';
+import {isNull} from '../../utils/index';
```
xexys: import {isNull} from '../../utils';

xexys (changes requested):

xexys (approved):

PR #1764, opened 2026-08-21
feat(Scheduling): add pool sort filter [YTFRONT-5930]

SimbiozizV:
https://nda.ya.ru/t/3VwRfvP27nsxGx
    ## Summary by Sourcery

Add pool operation sorting by FIFO, title, and FIFO index to the scheduling overview.

New Features:
- Add scheduling table sorting options for pool operations by FIFO order, title, and FIFO index.

Bug Fixes:
- Render column loading indicators only while loading is active.

Enhancements:
- Extend scheduling row sorting metadata and localized labels for title- and FIFO-based ordering.


packages/ui/src/ui/utils/scheduling/detailsTable.tsx
```diff
@@ -120,6 +120,31 @@ export const childTableItems = {
...
+        align: 'left',
+    },
+    name_fifo: {
+        sort(item: SchedulingRowData) {
+            if (item.type === 'operation') {
+                return item.fifoIndex;
+            }
+            return item.name;
+        },
+        get caption() {
```
xexys, line 143: There is no need to use a getter

SimbiozizV: In all `childTableItems`, i18n labels are already accessed via a getter — `name`, `type`, `user`, `FI`, and others. This is not a coincidence: `ytSetLang` calls `i18n.setLang` at runtime, and the getter reads the translation at the moment of access. A regular field will capture the string in the language of the module loading.

xexys (approved):

PR #1765, opened 2026-08-21
fix(UI): range date selector buttons [YTFRONT-5755]

SimbiozizV:
https://nda.ya.ru/t/0h4SfhLj7ntdQo
## Summary by Sourcery

Provide consistent range date selection and zoom controls across the UI.

New Features:
- Add a reusable range date selector with optional zoom controls and localized accessibility labels.

Bug Fixes:
- Fix range date selector zoom buttons across timeline and date-range consumers.

Enhancements:
- Centralize zoom behavior with support for minimum and maximum range durations.
- Update timeline integrations to use the shared range date selector component.

### DIFF
--- a/packages/ui/src/ui/components/RangeDateSelection/RangeDateSelection.tsx
+++ b/packages/ui/src/ui/components/RangeDateSelection/RangeDateSelection.tsx
@@ -1,0 +1,77 @@
+import React from 'react';
+
+import {Minus, Plus} from '@gravity-ui/icons';
+import {Button, Icon} from '@gravity-ui/uikit';
+import {
+    RangeDateSelection as RangeDateSelectionBase,
+    type RangeDateSelectionProps as RangeDateSelectionBaseProps,
+    type RangeValue,
+} from '@gravity-ui/date-components';
+
+import cn from 'bem-cn-lite';
+
+import {type DateTime, dateTimeParse} from '../../utils/date-utils';
+
+import i18n from './i18n';
+import {ZOOM_IN_FACTOR, ZOOM_OUT_FACTOR, calculateZoomedRange} from './zoom';
+
+import './RangeDateSelection.scss';
+
+const block = cn('yt-range-date-selection');
+
+export type RangeDateSelectionProps = Omit<RangeDateSelectionBaseProps, 'value'> & {
+    value: RangeValue<DateTime>;
+};
+
+export function RangeDateSelection({
+    className,
+    style,
+    hasScaleButtons,
+    scaleButtonsPosition = 'start',
+    ...props
+}: RangeDateSelectionProps) {
+    const {value, onUpdate, minDuration, maxDuration} = props;
+
+    const zoom = (factor: number) => {
+        const {from, to} = calculateZoomedRange({
+            from: value.start.valueOf(),
+            to: value.end.valueOf(),
+            factor,
+            minDuration,
+            maxDuration,
+        });
+
+        onUpdate?.({start: dateTimeParse(from)!, end: dateTimeParse(to)!});
+    };
+
+    const scaleButtons = hasScaleButtons ? (
+        <div className={block('buttons', {position: scaleButtonsPosition})}>
+            <Button
+                view="flat-secondary"
+                size="xs"
+                title={i18n('action_zoom-out')}
+                aria-label={i18n('action_zoom-out')}
+                onClick={() => zoom(ZOOM_OUT_FACTOR)}
+            >
+                <Icon data={Minus} />
+            </Button>
+            <Button
+                view="flat-secondary"
+                size="xs"
+                title={i18n('action_zoom-in')}
+                aria-label={i18n('action_zoom-in')}
+                onClick={() => zoom(ZOOM_IN_FACTOR)}
+            >
+                <Icon data={Plus} />
+            </Button>
+        </div>
+    ) : null;
+
+    return (
+        <div className={block(null, className)} style={style}>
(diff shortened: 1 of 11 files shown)


packages/ui/src/ui/components/RangeDateSelection/YTRangeDateSelection.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import './YTRangeDateSelection.scss';
+
+const block = cn('yt-range-date-selection');
+
+export type YTRangeDateSelectionProps = Omit<RangeDateSelectionProps, 'value'> & {
+    value: RangeValue<DateTime>;
+};
+
+export function YTRangeDateSelection({
```
xexys: It's not good that the component name doesn't match the folder name.

I suggest NOT renaming the component, but simply importing the gravity `RangeDateSelection` one under a different name (`RangeDateSelection as RangeDateSelectionBase`)

Also, if you do as I suggest, then you need to correct the css, remove the suffix yt.

PS:
The idea is that the components inside the project are kind of all about YT, and I don't want it to turn out that some of the components have a suffix and the other part doesn't.

packages/ui/src/ui/components/RangeDateSelection/RangeDateSelection.tsx
```diff
@@ -0,0 +1,77 @@
...
+    style,
+    hasScaleButtons,
+    scaleButtonsPosition = 'start',
+    ...props
+}: YTRangeDateSelectionProps) {
+    const {value, onUpdate, minDuration, maxDuration} = props;
+
+    const zoom = (factor: number) => {
+        const {from, to} = calculateZoomedRange({
+            from: value.start.valueOf(),
```
xexys, line 37: And what type of data is `value.start`? Is there really a need to use `valueOf`? Is it possible to replace it with something more explicit?

packages/ui/src/ui/components/RangeDateSelection/zoom.ts
```diff
@@ -0,0 +1,28 @@
+export const ZOOM_IN_FACTOR = 0.5;
+export const ZOOM_OUT_FACTOR = 2;
+
+export const FALLBACK_DURATION = 60 * 1000;
```
xexys: You can use a fancy separator :) `60_000`

packages/ui/src/ui/components/RangeDateSelection/zoom.ts
```diff
@@ -0,0 +1,28 @@
...
+    minDuration?: number;
+    maxDuration?: number;
+};
+
+export function calculateZoomedRange({
+    from,
+    to,
+    factor,
+    minDuration = 0,
+    maxDuration = Number.POSITIVE_INFINITY,
```
xexys: Maybe it's better to use `Number.MAX_SAFE_INTEGER` here?

packages/ui/src/ui/components/RangeDateSelection/zoom.ts
```diff
@@ -0,0 +1,28 @@
...
+}: ZoomedRangeParams) {
+    const duration = to - from;
+
+    const baseDuration = duration > 0 ? duration : Math.max(minDuration, FALLBACK_DURATION);
```
xexys: This check looks a bit strange. Are values less than 0 really assumed?

Maybe in this case it is more correct to throw an assert error?

packages/ui/src/ui/components/RangeDateSelection/RangeDateSelection.tsx
SimbiozizV: `value.start` / `value.end` is a `DateTime` from `@gravity-ui/date-utils` (a wrapper over `dayjs`); `.valueOf() ` returns a Unix timestamp in milliseconds.

packages/ui/src/ui/components/RangeDateSelection/RangeDateSelection.tsx
```diff
@@ -0,0 +1,77 @@
...
+import {type DateTime, dateTimeParse} from '../../utils/date-utils';
+
+import i18n from './i18n';
+import {ZOOM_IN_FACTOR, ZOOM_OUT_FACTOR, calculateZoomedRange} from './zoom';
+
+import './RangeDateSelection.scss';
+
+const block = cn('yt-range-date-selection');
+
+export type RangeDateSelectionComponentProps = Omit<RangeDateSelectionProps, 'value'> & {
```
xexys: Maybe `RangeDateSelectionProps` ?

packages/ui/src/ui/components/RangeDateSelection/RangeDateSelection.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+
+import {Minus, Plus} from '@gravity-ui/icons';
+import {Button, Icon} from '@gravity-ui/uikit';
+import {
+    RangeDateSelection as RangeDateSelectionBase,
+    type RangeDateSelectionProps,
```
xexys: type RangeDateSelectionProps as RangeDateSelectionBaseProps

packages/ui/src/ui/components/RangeDateSelection/RangeDateSelection.tsx
xexys: There’s also the `Override` utility from the `utility-types` module, which you can use precisely for such cases.

xexys (changes requested):

xexys (approved):

PR #1769, opened 2026-08-24
fix(Navigation/Table): download file as application/octet-stream [YTFRONT-5970]

xexys:
https://nda.ya.ru/t/_IgLbkr87oH9ZR
        ## Summary by Sourcery

Save table and query-result downloads as files instead of allowing browsers to open them in built-in viewers.

Bug Fixes:
- Ensure downloaded files are saved with the generic application/octet-stream MIME type so browsers do not intercept them for inline viewing.

Enhancements:
- Extract the download manager into a reusable shared component for navigation tables and query results.
- Centralize download format metadata, separator handling, filename extensions, and related utilities.
- Simplify download and clipboard handling around typed Blob responses and update end-to-end coverage for explicit downloads.

Tests:
- Update the static table download test to trigger and await the download request explicitly.


packages/ui/src/ui/pages/navigation/content/Table/DownloadManager/DownloadManager.tsx
```diff
@@ -887,40 +883,53 @@ export class DownloadManager extends React.Component<Props, State> {
...
 
-    renderModalConfirmButton(classNameConfirm: string) {
-        const {filename} = this.state;
+    renderModalConfirmButton = (classNameConfirm: string) => {
         const {url, error} = this.getDownloadLink();
+        const title = i18n('action_download');
 
         return (
-            <ConfirmButton
+            <Button
```
SimbiozizV: It looks like `filename` is being lost in this implementation.

packages/ui/src/ui/pages/navigation/content/Table/DownloadManager/DownloadManager.tsx
```diff
@@ -887,40 +883,53 @@ export class DownloadManager extends React.Component<Props, State> {
...
                 className={classNameConfirm}
-                filename={filename}
-                href={url}
-                title={i18n('action_download')}
+                size="m"
+                title={title}
                 disabled={Boolean(error)}
                 view="action"
+                href={url}
```
SimbiozizV: need `download` prop with filename

packages/ui/src/ui/pages/navigation/content/Table/DownloadManager/DownloadManager.tsx
```diff
@@ -1000,9 +1009,5 @@ const mapStateToProps = (state: RootState) => {
     };
 };
 
-const mapDispatchToProps = (dispatch: ThunkDispatch<any, any, any>) => ({
-    downloadFile: (filename: string, url: string) => dispatch(downloadFile(filename, url)),
```
SimbiozizV, line 1004: Without `downloadFile`, the toaster with the loading and progress disappears.

packages/ui/src/ui/components/DownloadManager/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+        .get(`${EXCEL_BASE_URL}/${cluster}/api/ready`)
+        .then(() => true)
+        .catch(() => false);
+};
+
+export const prepareSeparatorValue = (v?: string) => {
+    let res = v || '';
+    try {
+        res = JSON.parse(`"${v}"`);
+    } catch (e) {}
```
SimbiozizV, line 28: The parsing error is hidden from the user. This is not a blocker, but the cause of the crash will be difficult to understand later.

SimbiozizV (approved):

SimbiozizV (changes requested):
E2E problem

SimbiozizV (approved):

packages/ui/src/ui/components/DownloadManager/utils.ts
xexys: I agree, this is a bad pattern. This needs to be logged. But it’s just code transfer :/

PR #1787, opened 2026-09-02
feat(Accounts): load account details on demand [YTFRONT-5765]

SimbiozizV:
https://nda.ya.ru/t/2eGOBf2p7pQBnN
       ## Summary by Sourcery

Load account details lazily while retaining complete data for dashboard and general account views.

New Features:
- Load account metadata and detailed account data on demand based on the current account view.
- Load the selected account’s related hierarchy data for account editing and quota configuration.
- Provide loading and error states while account details and editor data are fetched.

Bug Fixes:
- Keep account list data and fetched details consistent when requests overlap, account context changes, or account edits occur.

Enhancements:
- Separate lightweight account-list loading from detailed resource, active-account, and editable-account requests.
- Preserve loaded account details when refreshing the account list and use the selected account’s data for general account views.

Tests:
- Add end-to-end coverage for parent and child account quota sources.
- Update account screenshots to wait for asynchronously loaded master-memory data.


packages/ui/src/ui/constants/accounts/accounts.ts
```diff
@@ -6,6 +6,10 @@ const ACCOUNTS_PREFIX = createPrefix('ACCOUNTS');
 export const FETCH_ACCOUNTS_RESOURCE = createActionTypes(
     ACCOUNTS_PREFIX + 'FETCH_ACCOUNTS_RESOURCE',
 );
+export const FETCH_ACCOUNTS_METADATA = createActionTypes(
```
ma-efremoff: The word “meta” can be misleading, as in most cases we use it to refer to data related to <MetaTable />.

packages/ui/src/ui/pages/accounts/tabs/statistic/AccountStatisticTab.tsx
```diff
@@ -3,24 +3,73 @@ import {useSelector} from '../../../../store/redux-hooks';
...
+import {YTApiId} from '../../../../rum/rum-wrap-api';
+import ypath from '../../../../common/thor/ypath';
+import {YTErrorBlock, type YTErrorBlockProps} from '../../../../containers/Block/Block';
+import {useGetQuery} from '../../../../store/api/yt/get';
+
+interface AccountTree {
+    [name: string]: AccountTree;
+}
+
+function collectAccountNames(account: string, tree?: AccountTree): Array<string> {
```
ma-efremoff, line 20: it looks like an existing helper fieldTreeForEach

packages/ui/src/ui/pages/accounts/AccountsSuggest.tsx
```diff
@@ -115,14 +115,23 @@ export function AccountsSuggestWithLoading(
         items?: Array<string>;
         error?: YTError;
     }>({items: []});
-    React.useMemo(() => {
+    React.useEffect(() => {
+        let active = true;
```
ma-efremoff: isn't it better to name it `mounted`?

packages/ui/src/ui/pages/accounts/Accounts/Accounts.tsx
```diff
@@ -171,7 +171,10 @@ const connector = connect(mapStateToProps);
 export default connector(Accounts);
 
 function AccountsRumMeasure() {
-    const isFinalStatus = useSelector(selectAccountsIsFinalLoadingStatus);
+    const accountsFinalStatus = useSelector(selectAccountsIsFinalLoadingStatus);
+    const {pathname} = useLocation();
+    const isGeneralTab = pathname.endsWith('/' + AccountsTab.GENERAL);
```
ma-efremoff: Is there a way to avoid such check here?

packages/ui/src/ui/pages/accounts/Accounts/AccountsUpdater.tsx
```diff
@@ -1,19 +1,49 @@
...
 import {selectAccountsEditCounter} from '../../../store/selectors/accounts/accounts-ts';
+import {AccountsTab} from '../../../constants/accounts/accounts';
 import {useUpdater} from '../../../hooks/use-updater';
 
-export default function AccountsUpdater() {
+type Props = {
+    loadAllAccountDetails?: boolean;
+};
+
+export default function AccountsUpdater({loadAllAccountDetails = false}: Props) {
```
ma-efremoff, line 14: can we incapsulate all the logic in fetchAccounts?

packages/ui/src/ui/pages/accounts/tabs/general/AccountsGeneralTab.js
```diff
@@ -702,13 +704,27 @@ class AccountsGeneralTab extends Component {
...
+                ...baseTableProps.columns,
+                items: {
+                    ...baseTableProps.columns.items,
+                    name: {
+                        ...baseTableProps.columns.items.name,
+                        captionTail: metadataFetching ? <Loader visible /> : undefined,
+                    },
+                },
+            },
+        };
```
ma-efremoff: isn't it better to pass `loader` to getTableProps?

packages/ui/src/ui/pages/accounts/tabs/statistic/AccountStatisticTab.tsx
```diff
@@ -3,24 +3,73 @@ import {useSelector} from '../../../../store/redux-hooks';
...
+            result.push(name);
+            visit(nested);
+        });
+    }
+
+    visit(tree);
+    return result;
+}
+
+function useAccountSubtreeNames(account: string, accountPath?: string) {
```
ma-efremoff, line 31: Isn't it better to move the code to CustomAccountStatistics?

packages/ui/src/ui/pages/accounts/Accounts/Accounts.tsx
SimbiozizV: We still need to distinguish the General tab here because the full accounts request is only started for that tab, while AccountsRumMeasure is mounted for the whole Accounts page. Otherwise the measure would wait for a request that is never started on the lazy-loaded tabs.
I moved the pathname handling to the shared isAccountsGeneralPath helper, so the updater and RUM use exactly the same check.

packages/ui/src/ui/pages/accounts/tabs/statistic/AccountStatisticTab.tsx
SimbiozizV: I’d prefer to keep this logic in AccountStatisticTab. The current UIFactory contract provides accountSubtreeAllNames to the custom statistics component, so the tab is responsible for loading and preparing that data. Moving it into CustomAccountStatistics would change the contract and couple the custom component to the YT API. We can extract useAccountSubtreeNames into a separate hook if we want to make the tab itself smaller.

ma-efremoff (approved):

PR #1804, opened 2026-09-11
feat(MetaTable): render Nirvana links with inline errors [YTFRONT-6026]

xexys:
https://nda.ya.ru/t/pzix5ub87qcwYe
 ## Summary by Sourcery

Enable inline error handling for Nirvana links rendered in MetaTable.

New Features:
- Render Nirvana block URLs in MetaTable through the shared Markdown component with inline error display.

Bug Fixes:
- Preserve user-visible errors when Nirvana link Markdown rendering fails instead of silently showing empty content.

Enhancements:
- Refactor Markdown transformation into a reusable hook and add configurable block or inline error presentation.
- Update the YTsaurus components dependency to support the Nirvana block URL renderer.


packages/ui/src/ui/components/Markdown/MarkdownFallback.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import {useMarkdown} from '../../hooks/useMarkdown';
+
+const block = cn('yt-markdown');
+
+type Props = {
+    text: string;
+};
+
+export const MarkdownFallback = ({text}: Props) => {
```
ma-efremoff, line 12: **not a blocker**: it’s more accurate to talk about the default implementation rather than a fallback.

ma-efremoff (approved):

PR #1819, opened 2026-09-17
fix(Navigation): use real paths for queue and consumer metrics [YTFRONT-5853]

xexys:
https://nda.ya.ru/t/v-c1qHdM7rPntH
## Summary by Sourcery

Use real navigation paths for queue and consumer metrics while informing users when the displayed context differs from the actual path.

New Features:
- Add an informational notice linking users from virtual navigation paths to the corresponding real queue or consumer path.

Bug Fixes:
- Use real queue and consumer paths when loading their metrics.

Enhancements:
- Improve queue and consumer page layout to accommodate path notices and consistently group metadata and content.


SimbiozizV (approved):

SimbiozizV (changes requested):
I saw that the task would be up for discussion.

SimbiozizV (approved):

PR #1823, opened 2026-09-18
test: support external screenshot baselines [YTFRONT-5912]

ma-efremoff:
## Summary by Sourcery

Enable UI end-to-end tests to use screenshot baselines stored outside the test container.

New Features:
- Support configuring an external Playwright screenshot baseline directory for UI end-to-end tests.

Enhancements:
- Validate and bind-mount external screenshot directories when running tests in containers.

Build:
- Configure dependency installation to disable legacy peer dependency handling.


packages/ui/tests/playwright.config.ts
```diff
@@ -24,6 +24,7 @@ function storageState() {
 export default defineConfig({
     timeout: 10000,
     testDir: E2E_TEST_DIR,
+    snapshotDir: process.env.E2E_SNAPSHOT_DIR,
```
xexys: It’s better to do the destructuring as for other `process.env` fileds.

xexys (changes requested):

